PR #10, opened 2022-05-10
Added Ackee Template

Ponkhy:


deiucanta:
1. Let's remove the MongoDB password field.
2. I'm afraid people will be confused about this instruction "Needs to be copied from the 'Environment' tab". You can make the Ackee password field required and not generate a random one.

A better experience would be to have a message after the app is created with instructions on how to access the newly created app. I'll create a separate issue to discuss this.

Ponkhy:
Also removed the fields here.

templates/ackee.ts
```diff
@@ -0,0 +1,86 @@
...
+        port: 3000,
+        secure: true,
+      },
+      domains: [{ name: domain }],
+    };
+
+    const mongoService: MongoService = {
+      projectName,
+      serviceName: mongoServiceName,
+      image: "mongo:4.4.13",
```
deiucanta, line 75: Do you have a special reason for setting the version like this? I think it's better to keep `mongo:4`. It will still get the latest `4.*` but it's pinned to the major version.

Ponkhy: Yess, Just noticed it too, thanks!
Fixed in [709012d](https://github.com/easypanel-io/templates/pull/10/commits/709012df98cc342bc90d10516e43ff17eb887362)

PR #11, opened 2022-05-10
Added Gitea Template

Ponkhy:


Ponkhy:
Jup, you're right, i did remove the field now.

templates/gitea.ts
```diff
@@ -0,0 +1,122 @@
...
+      },
+      domains: [{ name: domain }],
+      volumes: [
+        {
+          type: "volume",
+          source: "gitea",
+          target: "/data",
+        },
+        {
+          type: "volume",
```
deiucanta, line 83: this should be `type: "bind"`

templates/gitea.ts
```diff
@@ -0,0 +1,122 @@
...
+          source: "/etc/timezone",
+          target: "/etc/timezone:ro",
+        },
+        {
+          type: "volume",
```
deiucanta, line 88: this should be `type: "bind"`

templates/gitea.ts
```diff
@@ -0,0 +1,122 @@
...
+      password: databasePassword,
+    };
+
+    const mysqlService: MySQLService = {
+      projectName,
+      serviceName: databaseServiceName,
+      password: databasePassword,
+    };
+
+    let databaseService = databaseType == "postgres" ? postgresService : mysqlService;
```
deiucanta, line 113: use `===`

templates/gitea.ts
```diff
@@ -0,0 +1,122 @@
...
+    },
+  } as const,
+  generate({
+    projectName,
+    domain,
+    appServiceName,
+    databaseType,
+    databaseServiceName,
+  }) {
+    let databasePassword = randomPassword();
```
deiucanta, line 50: use const instead of let

Ponkhy:
Okay, Should be fixed now

PR #12, opened 2022-05-10
Added Directus Template

Ponkhy:


deiucanta:
Wow! - thanks for the PR.

I'm wondering what fields are not mandatory and can be removed to simplify the form.

1. Can we remove the admin email/password? If Directus has a setup page, maybe you can configure the admin credentials there.
2. App key and app secret can be removed from the form as I suppose nobody wants to customize them. You will still generate them randomly and pass them into the environment, but don't make them configurable.
3. Database password and Redis password should also be removed from the form.

Ponkhy:
Okay, i did remove the app key and secret field from the form, same for the redis and database.
Unfortunately it looks like the admin login is required for the Docker image, but i made the password field mandatory.

templates/directus.ts
```diff
@@ -0,0 +1,141 @@
...
+    appServiceName,
+    adminEmail,
+    adminPassword = randomPassword(),
+    databaseType,
+    databaseServiceName,
+    appKey = randomString(48),
+    appSecret = randomString(48),
+    redisServiceName,
+  }) {
+    let databasePassword = randomPassword();
```
deiucanta, line 71: use const

templates/directus.ts
```diff
@@ -0,0 +1,141 @@
...
+        title: "Domain",
+      },
+      appServiceName: {
+        type: "string",
+        title: "App Service Name",
+        default: "directus",
+      },
+      adminEmail: {
+        type: "string",
+        title: "Admin E-Mail",
```
deiucanta, line 32: hope I'm not to picky, but can we have `Email` instead of `E-Mail`?

templates/directus.ts
Ponkhy: Fixed in [27974ce](https://github.com/easypanel-io/templates/pull/12/commits/27974cef460b51c12f16f7d0fbe0ac4a2912a67c)

templates/directus.ts
```diff
@@ -0,0 +1,141 @@
...
+        secure: true,
+      },
+      domains: [{ name: domain }],
+      volumes: [
+        {
+          type: "volume",
+          source: "uploads",
+          target: "/directus/uploads",
+        },
+      ],
```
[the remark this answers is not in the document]
Ponkhy: Yes! Thank you, i didn't see that one.
Fixed in [27974ce](https://github.com/easypanel-io/templates/pull/12/commits/27974cef460b51c12f16f7d0fbe0ac4a2912a67c)

Ponkhy:
Okay, Should be fixed here now too :)

PR #15, opened 2022-05-12
Added Statping-ng Template

Ponkhy:


deiucanta:
For a service like this, I would use SQLite since it doesn't have a lot of traffic. What do you think, should we support multiple database types or should we default to SQLite?

Ponkhy:
I think we should support multiple database types and just set the default to SQLite, so ppl still have the option in case they want something different.

PR #16, opened 2022-05-12
Added Wiki.js Template

Ponkhy:


deiucanta:
Should we add SQLite as a DB option?

Ponkhy:
Would do the same as mentioned [here](https://github.com/easypanel-io/templates/pull/15#issuecomment-1124875239).

Ponkhy:
Looks like there is a problem with SQLite and Wiki.js right now.
The beta of Wiki.js 3 also came out this month and will support only Postres anyway.
So i'd say we stay with the current config.

templates/wikijs.ts
```diff
@@ -0,0 +1,98 @@
...
+        port: 3000,
+        secure: true,
+      },
+      domains: [{ name: domain }],
+    };
+
+    const postgresService: PostgresService = {
+      projectName,
+      serviceName: databaseServiceName,
+      image: "postgres:alpine",
```
deiucanta, line 79: do you need to change the default image for postgres?

templates/wikijs.ts
```diff
@@ -0,0 +1,98 @@
...
+        title: "App Service Name",
+        default: "wikijs",
+      },
+      databaseType: {
+        type: "string",
+        title: "Database Type",
+        oneOf: [
+          { enum: ["postgres"], title: "Postgres" },
+          { enum: ["mysql"], title: "MySQL" },
+        ],
```
deiucanta, line 34: you can add a `default` prop for `postgres`

deiucanta:
> Looks like there is a problem with SQLite and Wiki.js right now. The beta of Wiki.js 3 also came out this month and will support only Postres anyway. So i'd say we stay with the current config.

Sounds good.

templates/wikijs.ts
```diff
@@ -0,0 +1,106 @@
...
+    appServiceName,
+    databaseType,
+    databaseServiceName,
+  }) {
+    const services: Services = [];
+    const databasePassword = randomPassword();
+    const databaseUsername = databaseType === "postgres" ? "postgres" : "mysql";
+    const databasePort = databaseType === "postgres" ? "5432" : "3306";
+
+    services.push({
```
deiucanta, line 60: I think you forgot to specify the service type

Ponkhy: Haha, Thanks, was a little too quick there.

PR #32, opened 2022-05-16
Added MeTube Template

Ponkhy:


templates/metube.ts
```diff
@@ -0,0 +1,71 @@
...
+        },
+        proxy: {
+          port: 8081,
+          secure: true,
+        },
+        domains: [{ name: domain }],
+        volumes: [
+          {
+            type: "volume",
+            source: downloadPath,
```
deiucanta, line 62: I think we should not make this configurable. It should be "downloads" by default.

Ponkhy: Normally i would agree, but in this case it's pretty likely that users want to change it and at the end this is easier than explaining how to create a volume for this image afterwards.

deiucanta: Makes sense. Then let's name it "Downloads Volume Name" (maybe). Or, if you want to keep it as a path, then the volume type should be `bind` not `volume`

deiucanta: @Ponkhy I want to merge this PR. Can you please change "Download Path" to "Downloads Volume Name"?

Ponkhy: Sorry for the delay, it's renamed now.

PR #64, opened 2022-06-22
Add PsiTransfer Template

marktopper:
https://hub.docker.com/r/psitrax/psitransfer


marktopper:
**Issue:**
The data volume needs UID 1000
```chown -R 1000 /data```

Any idea how this can be resolved so the template can work fully with no additional requirements.

PR #82, opened 2022-08-06
Fix required field for Drone Template

ivanonpc-22:


ivanonpc-22:
Should fix issue when generating json.

PR #91, opened 2022-08-18
add dashy template

bedeoan:


templates/dashy/index.ts
```diff
@@ -0,0 +1,57 @@
+import { createTemplate, Services } from "~templates-utils";
+
+export default createTemplate({
+  name: "Dashy",
+  meta: {
+    description:
+      "Dashy helps you organize your self-hosted services by making them accessible from a single place      ",
```
deiucanta, line 7: you have unnecessary empty spaces at the end

templates/dashy/index.ts
```diff
@@ -0,0 +1,57 @@
...
+    instructions:
+      "After deploying the project, it may take a while before the app is available(5 min)",
```
deiucanta, line 9: add a space before the `(` and use `minutes` instead of `min`

templates/dashy/index.ts
```diff
@@ -0,0 +1,57 @@
...
+    const services: Services = [];
+
+    services.push({
+      type: "app",
+      data: {
+        projectName,
+        serviceName,
+        source: { type: "image", image: "lissy93/dashy" },
+        domains: [{ name: domain }],
+        proxy: { port: 80, secure: true },
```
deiucanta, line 50: don't set `secure: true`, as it is the default

templates/dashy/index.ts
```diff
@@ -0,0 +1,57 @@
...
+        deploy: { replicas: 1, command: null, zeroDowntime: true },
```
deiucanta, line 51: these `deploy` options are the default ones, you can remove them

templates/dashy/index.ts
```diff
@@ -0,0 +1,57 @@
...
+        env: [`NODE_ENV=production`].join("\n"),
```
deiucanta, line 52: if there is a single line in the env, you can use a simple string (no need for array and join)

templates/dashy/index.ts
```diff
@@ -0,0 +1,57 @@
...
+      },
+    });
+    return { services };
```
deiucanta, line 55: add an empty line before the return statement

deiucanta (changes requested):

PR #98, opened 2022-10-12
Mattermost template: Ability to define docker image version

fodurrr:
Hi,

I just modified the Mattermost template to be able to define the docker image before first pull.

I tried with json schema and it works fine.


deiucanta:
Thanks @fodurrr for your contribution!

PR #109, opened 2022-11-01
Fix changedetection logo

Supernova3339:
I noticed that .ico format is NOT supported, changing to a png.


Supernova3339:
wait hold on
lemme close this

PR #110, opened 2022-11-01
Fix changedetection logo in Easypanel 🎉

Supernova3339:
Noticed that the icon format .ico is not supported within easypanel. Changing to a .png format to fix.


Supernova3339:
Oh yeah I accidentally put the wiki as the description in SimpleTorrent, fix in 8 hours or sooner :)

PR #114, opened 2022-11-02
Add Flarum template

d8vjork:


deiucanta:
Hey @d8vjork, thanks for this PR! I see you have separate fields for the database settings. Can we simplify that like the other templates and simply don't ask for those fields? This would mean that an external DB cannot be configured, but I think most people will use an internal DB. If someone wants to use external DB, they can change the env manually after creating the service.

d8vjork:
@deiucanta Ah, for me the DB fields makes sense as I use RDS (a managed DB) and so I don't want Easypanel to create a DB instance on the server

Will it be possible to just let the user decide to not use any prebuilt DB instead?

deiucanta:
@d8vjork then it makes sense to keep them :) thanks!

PR #115, opened 2022-11-02
HedgeDoc Quick Fix

Supernova3339:
HedgeDoc recently introduced a content security policy, which has the enviroment variable CMD_DOMAIN.

This quick little PR fixes that. Your welcome :)


deiucanta:
Thanks for the PR. The formatting was not right. Also, it seems you edited the files in Github and haven't run the `build-templates` script. I implemented this directly in `main` because I could not merge it.

PR #116, opened 2022-11-04
Add Flarum to Easypanel 🎉

Supernova3339:
Shoutout to @d8vjork for the template, modified it a bit and fixed some issues that will break the forum when deployed.


deiucanta:
I merged this offline because of conflicts.

PR #118, opened 2022-11-06
Add MailCrab to Easypanel 🎉

Supernova3339:
I don't wanna rename the branch but it is MailCrab!
It was uploaded a few days ago to docker, so the only version available is **latest**
Compatable with `AMD` and `ARM` CPUs, so won't have any problems with deployment.


deiucanta:
Thanks for the PR!

Supernova3339:
> Thanks for the PR!

No worries! Anything I find not compatible with AMP CPUs, I want to provide a good alternate solution.

PR #120, opened 2022-11-08
Add RabbitMQ template

d8vjork:
As mentioned on the title, **but also fixing the Mailhog logo's ratio**

### Side notes

- As advised by RabbitMQ official documentation, the template encourage the user to use a memory limit which is defaulted by a static unit (as **I don't know if using the node's OS will be a limitation here or not**)

- Ports are toggled depending on the Management UI, if this checkbox is enabled the image is enforced to be the `-management` which contains this plugin by default. So default port is Management UI and queue's port as additional port


deiucanta:
Merged it locally in order to fix the conflicts.

PR #133, opened 2022-11-22
Add Home Assistant to Easypanel 🎉

Supernova3339:
This adds Home Assistant to Easypanel


migu2k:
The template requires certain configurations to be added so that it can be used with Traefik as the reverse proxy. For reference see: https://community.home-assistant.io/t/home-assistant-400-bad-request-docker-proxy-solution/322163

migu2k:
Error log from fresh deploy:
`2022-11-28 11:57:49.149 ERROR (MainThread) [homeassistant.components.http.forwarded] A request from a reverse proxy was received from 10.0.1.6, but your HTTP integration is not set-up for reverse proxies`

PR #142, opened 2022-11-30
[+] Templates ready to go!

theh2so4:
[+] Sonarr added as a template, already tested!


migu2k:
Don't forget the entries in the /templates/index.ts file :)

deiucanta:
@theh2so4 thanks for the PR!

Please add fixed versions for the docker images instead of `latest`. Sometimes the newer versions break the templates and we want to avoid that. 

Also, I see the logo and the screenshots were not picked up by the template builder. If you run the build command again, the `meta.ts` files should have the logo and screenshot file names.

Supernova3339:
I don't think it's a good idea to do homer. I did it myself, then I realized there isn't a GUI editor. You need to install a 3rd party tool to even consider customizing it. If it was possible, I would've added it, last month.

PR #143, opened 2022-11-30
[+] Added OpenSpeedTest

migu2k:


deiucanta:
@migu2k thanks for the PR!

Please add fixed versions for the docker images instead of latest. Sometimes the newer versions break the templates and we want to avoid that.

Also, I see the logo and the screenshots were not picked up by the template builder. If you run the `yarn build-templates` command again, the `meta.ts` files should have the logo and screenshot file names.

migu2k:
> Please add fixed versions for the docker images instead of latest. Sometimes the newer versions break the templates and we want to avoid that.

I've done some checking, and it seems like they only release on the latest version and don't offer older versions to be deployed. I've also tried to deploy older versions, though none of them succeeded and all resulted in errors. 

So what do we do now?

PS. I also reran yarn build-templates so that should be correct now :)

deiucanta:
@migu2k makes sense! I merged it ;)

PR #150, opened 2022-12-12
Add Mail to Easypanel  🎉

Supernova3339:
This adds a mailserver, via using SMTP relay.


deiucanta:
This is not a full mail server, but a mail relay server. I'm afraid people will be confused by it.

Supernova3339:
I'll change it to `SMTP Relay`
Does that work?

deiucanta:
I started to merge it locally but it doesn't seem to work. Port 25 should not be proxied since it's for SMTP, not HTTP.

deiucanta:
@Supernova3339 I don't think that this is a good fit for Easypanel templates. Also, it's incomplete. Closing for now. Feel free to open again.

Supernova3339:
hi, this is meant to be integrated with services such as plausible for SMTP
relay. It is NOT meant as an actual server!
If changed to "smtp relay", would that work better? Waiting for your reply
:)

PR #157, opened 2022-12-20
feat: add libretranslate template

kaname-png:


templates/libretranslate/meta.yaml
```diff
@@ -0,0 +1,68 @@
...
+      default: 100
+    googleAnalytics:
+      type: string
+      title: Google Analytics
+      default: ""
+    suggestions:
+      type: boolean
+      title: Translation suggestions
+      default: false
+    webUI:
```
deiucanta, line 62: I think this should be called `disableWebUI` instead.

templates/libretranslate/meta.yaml
```diff
@@ -0,0 +1,68 @@
...
+      type: boolean
+      title: Disable Web UI
+      default: false      
+    
+    
+
```
deiucanta, line 68: Also, remove unused empty lines.

PR #162, opened 2022-12-22
Added Memos to Easypanel 🎉

DrMxrcy:


Supernova3339:
no joke I never added this because I couldn't get it to work at all

DrMxrcy:
> no joke I never added this because I couldn't get it to work at all

I got it running, Tested it right through EasyPanel. Runs pretty good

PR #163, opened 2022-12-22
Add Blog to Easypanel 🎉

Supernova3339:
Facebook go 💀


templates/blog/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+    blogName:
+      type: string
+      title: Blog Name
+    blogNick:
+      type: string
+      title: Username
+    blogPass: 
+      type: string
+      title: Password
+    blogLang:
```
deiucanta, line 47: I think you should add this to the `required` array.

templates/blog/meta.ts
```diff
@@ -0,0 +1,65 @@
...
+        default: "m1k1o/blog:latest",
+      },
+      blogTitle: { type: "string", title: "Blog Title" },
+      blogName: { type: "string", title: "Blog Name" },
+      blogNick: { type: "string", title: "Username" },
+      blogPass: { type: "string", title: "Password" },
+      blogLang: { type: "string", title: "Language", default: "EN" },
+    },
+  },
+  logo: "logo.png",
```
deiucanta, line 43: The logo is the FB favicon from what I can see. If it doesn't have a real one, just remove it and a placeholder will be used by Easypanel.

Supernova3339: This is the official logo supplied in the github repo, hence what I used.

deiucanta:
@Supernova3339 I don't think that this is a good fit for Easypanel templates. We're looking to keep the base of templates on a higher standard.

PR #175, opened 2023-01-11
Add Odoo to EasyPanel!

DrMxrcy:


templates/odoo/index.ts
```diff
@@ -0,0 +1,67 @@
...
+        {
+          type: "volume",
+          name: "odoo-addons",
+          mountPath: "/mnt/extra-addons",
+        },
+      ],
+    },
+  });
+  services.push({
+    type: "app",
```
deiucanta, line 50: Any reason for not using the dedicated service for postgres?

deiucanta:
The `yarn.lock` file changes keep coming back. In order to not have it in the PR, you need to not push the changes. In this PR you deleted the file and pushed that. It's still a change with the file. You can try to undo your commit and revert any changes to the lock file. If you have issues, I can do it on my local machine and merge it there.

templates/odoo/index.ts
DrMxrcy: The reason is becuase Odoo would not work with the postgres username, and I saw no way to change it with the dedicated service

deiucanta:
@DrMxrcy I merged it, but please confirm that this requires a custom postgres service and not the "native" one which comes with Easypanel.

DrMxrcy:
If you have any more questions let me know!

PR #176, opened 2023-01-11
Add SuiteCRM To EasyPanel!

DrMxrcy:


templates/suitecrm/index.ts
```diff
@@ -0,0 +1,55 @@
...
+  randomPassword,
+  randomString,
+  Services,
+} from "~templates-utils";
+import { Input } from "./meta";
+
+export function generate(input: Input): Output {
+  const services: Services = [];
+  const mariaPassword = randomPassword();
+  const appKey = Buffer.from(randomString(32)).toString("base64");
```
deiucanta, line 12: Isn't `randomString` enough here?

templates/suitecrm/meta.yaml
```diff
@@ -0,0 +1,35 @@
+name: SuiteCRM
+description: SuiteCRM is a completely open source, enterprise-grade Customer Relationship Management (CRM) application.
+instructions: null
+changeLog:
+  - date: 2023-1-12
+    description: first release
+links:
+  - label: Github
+    url: https://github.com/salesagility/SuiteCRM
```
deiucanta, line 9: Can you please add the website? https://suitecrm.com/. You can add the docs as well: https://docs.suitecrm.com/

templates/suitecrm/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+      title: App Service Name
+      default: suitecrm
+    appServiceImage:
+      type: string
+      title: App Service Image
+      default: bitnami/suitecrm:8
+    databaseServiceName:
+      type: string
+      title: Database Service Name
+      default: suitecrm-roadmap-db
```
deiucanta, line 35: Can we make this name simply `suitecrm-db`?

deiucanta:
Here is a better logo

templates/suitecrm/index.ts
Supernova3339: Checked myself, needs to be this specific to work

PR #179, opened 2023-01-21
Add Zipline to EasyPanel!

DrMxrcy:


deiucanta:
The logo should be a square image. Can you please replace it? Other than that, this PR looks good.

PR #182, opened 2023-01-22
Add Chatwoot to Easypanel 🎉 

Supernova3339:
Finally, a drop-in live chat solution you don't have to pay for.


deiucanta:
Great one, thanks @Supernova3339!

jeanjonathas:
Chatwoot has a system that sends a URL to customers to evaluate the service. This URL in Chatwoot with Easypanel is broken. It is sent as https:///survey/responses/. The domain configured in .env was not supposed to appear like this https://my/ domain here/survey/responses/
I installed the same version of Chatwoot on a VPS with only Chatwoot and the link works normally.

PR #207, opened 2023-01-28
Add LogChimp to EasyPanel!

DrMxrcy:
**Note:** The image only has a working "latest" tag. When I get around to it, I can build a versioned one.


templates/logchimp/meta.yaml
```diff
@@ -0,0 +1,39 @@
+name: LogChimp
+description: Build better products with customer feedback.
+instructions: null;
```
deiucanta, line 3: I'm not sure if this is a correct value for instructions. Leave it empty if there are no instructions.

DrMxrcy: Whoops, sorry. Put that there by accident!

DrMxrcy:
Closed due to old image, and a lack of an official image.

Supernova3339:
> Closed due to old image, and a lack of an official image.
ghcr.io/logchimp/logchimp
no versioning so unofficial is fine

PR #235, opened 2023-02-24
fixed rallly links

AgentGeek:
Rallly links were incorrect and pointing to fider.io docs and website.


deiucanta:
Thanks for your contribution!

PR #245, opened 2023-02-28
Add MinIO to Easypanel

Supernova3339:
This is going to make things a lot easier


deiucanta:
I'm not sure this is going to work. Minio has two ports: one for the API and one for UI and we need to expose both of them. Right now we can only proxy one. I need to change they way domains/proxies work in Easypanel in order to accommodate this.

Supernova3339:
> I'm not sure this is going to work. Minio has two ports: one for the API and one for UI and we need to expose both of them. Right now we can only proxy one. I need to change they way domains/proxies work in Easypanel in order to accommodate this.

The API can be accessed internally and externally by providing port 9000
Check typebot or outline for an example

deiucanta:
That's true, but we need SSL support for the API as well.

loan-talvat:
Caprover makes two containers in its template
- a principal with the image: minio.domain.ltd
- an nginx reverse proxy on port 9000: s3.domain.ltd

We run Minio on Easypanel by modifying the Traefik config to avoid a reverse proxy:
image

Supernova3339:
> Caprover makes two containers in its template
> 
> * a principal with the image: minio.domain.ltd
> * an nginx reverse proxy on port 9000: s3.domain.ltd
> 
> We run Minio on Easypanel by modifying the Traefik config to avoid a reverse proxy: image

Once an update that makes this possible is released, this will be merged and added along with it.

loan-talvat:
> Once an update that makes this possible is released, this will be merged and added along with it.

https://feedback.easypanel.io/feature-requests/p/redirects
https://feedback.easypanel.io/feature-requests/p/manage-reverse-proxy

If easypanel can effectively manage redirections and reverse proxy, it can be a clean solution!

Supernova3339:
That may be true. However, both containers can be separated. If I can
figure that out, it will get PRd.
No clues really, so if you've got the console deployed, please share your
ENV, storage, and image setup :)

deiucanta:
let's make this as two services, one for api and one for admin ui

Supernova3339:
> let's make this as two services, one for api and one for admin ui

If I can get his setup shared here, it can be done!

ikbenignace:
Since you can have mulitple domains pointed to different ports, this should be possible

Supernova3339:
> Since you can have mulitple domains pointed to different ports now, this should be possible

I am currently looking into an implementation method as of now, will soon undraft this and have it ready for next release

PR #262, opened 2023-03-14
Add YAO to Easypanel 🎉 

Supernova3339:


templates/yao/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - appServiceImage
+    - databaseServiceName
+  properties:
+    projectName:
+      type: string
+      title: Project Name
+    appServiceName:
+      type: string
+      title: App Service Name
+      default: wordpress
```
Ponkhy, line 31:
```suggestion
      default: yao
```

Ponkhy (approved):
LGTM

templates/yao/meta.yaml
```diff
@@ -0,0 +1,39 @@
+name: YAO
+description: YAO is an open-source application engine,
+instructions: Use xiang@iqka.com:A123456p+ to login at /xiang/login/admin . Use amd64 instead of arm64 if you use a non ARM based system.
```
deiucanta, line 3: is there a way to change the default credentials using env vars?

Supernova3339: this is hardcoded sadly and there is no way to change it as of now.

PR #266, opened 2023-03-15
Add Lynx to Easypanel 🎉 

Supernova3339:


templates/lynx/meta.yaml
```diff
@@ -0,0 +1,41 @@
+name: Lynx
+description: A fullstack application using the MEVN stack to shorten your URLs.
+instructions: null
```
Ponkhy, line 3: We should maybe add a instructions like "Registration is disabled by default, but the first registration will always be allowed."

PR #267, opened 2023-03-15
Add pgweb to Easypanel 🎉 

Supernova3339:


templates/pgweb/meta.yaml
```diff
@@ -0,0 +1,32 @@
+name: pgweb
+description: Cross-platform client for PostgreSQL databases
+instructions: null
+changeLog:
+  - date: 2023-3-15
+    description: first release
+links:
```
Ponkhy, line 7: You could also add the Demo page `https://pgweb-demo.fly.dev`.

Ponkhy:
Maybe add some fields for the `DATABASE_URL` env var, so people can fix their database connection?

Supernova3339:
That's not needed at all, it's for autoconnection

templates/pgweb/meta.yaml
Supernova3339: not implemented

Ponkhy (approved):
LGTM

PR #268, opened 2023-03-15
Add Grafana to Easypanel 🎉 

Supernova3339:


berksmbl (approved):
Everything seems fine.

templates/grafana/index.ts
```diff
@@ -0,0 +1,37 @@
+import {
+  Output,
+  Services
+} from "~templates-utils";
```
Ponkhy, line 4:
```suggestion
import { Output, Services } from "~templates-utils";
```

templates/grafana/meta.yaml
```diff
@@ -0,0 +1,34 @@
+name: Grafana
+description: The open and composable observability and data visualization platform
+instructions: Login with admin:admin
+changeLog:
+  - date: 2023-3-15
+    description: first release
+links:
```
Ponkhy, line 7: You could also add the Demo page `https://play.grafana.org/`.

templates/grafana/index.ts
berksmbl: I totally agree, I recommend using the [prettier](https://prettier.io/) app.

Supernova3339:
Andrei runs it whenever PRs get merged, I don't touch it as it messes up my
flow

templates/grafana/meta.yaml
Supernova3339: not implemented

Ponkhy (approved):
LGTM

PR #269, opened 2023-03-16
Add Hasura to Easypanel

amitoo7:
Please help me add this Hasura template to EasyPanel. I have already tested it on the Easypanel instance and it's working fine.


templates/hasura/meta.yaml
```diff
@@ -0,0 +1,73 @@
...
+        - enum:
+            - "true"
+          title: "true"
+        - enum:
+            - "false"
+          title: "false"
+    hasuraDomain:
+      type: string
+      title: Hasura Domain name
+      default: Hasura domain
```
Ponkhy, line 73: I'd probably recommend to leave the default domain value empty, so the user definitely fills it.
```suggestion

```

amitoo7: Thanks, I think you are right

Ponkhy (approved):
LGTM

PR #270, opened 2023-03-16
add: Joplin

berksmbl:


templates/joplin/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+    - appServiceImage
+    - databaseServiceName
+  properties:
+    projectName:
+      type: string
+      title: Project Name
+    domain:
+      type: string
+      title: Domain
+    secure:
```
deiucanta, line 33: let's not expose this config, and make it secure by default

berksmbl: I added this option later because: If I do not specify https or http to the domain variable in the APP ENV, joplin does not allow connection.

If I specify https, it does not accept incoming connections over http.

Ponkhy: Shouldn't be a problem since Easypanel is requesting a certificate etc.
If people want http they're still able to change it afterwards.

berksmbl: image

berksmbl: I have presented this option considering the possibility that users will not use SSL in the domain they have specified since they will not know the default domain name by heart.

deiucanta: > it does not accept incoming connections over http

I think Easypanel would redirect that to https automatically. Right?

charliemoss:
Works great!

Just to mention. When changing the admin email, it says it has emailed a confirmation link to the new address. This email never arrived. However, that email could be found in the Admin/Emails section, from where the confirmation like could be clicked.

Supernova3339:
You'll need to configure an email server to utilize this feature

PR #275, opened 2023-03-16
add: Photoview

berksmbl:


Ponkhy:
Hey, I think the `api_cache:/app/cache` volume should be added and maybe create another user input for the photos path?

berksmbl:
> maybe create another user input for the photos path?
I thought about it but realized it might be confusing.
Because in the first step of the installation, this path value will be requested again.
And it can have a different variant for each user.
I think it can be easily mounted via Easypanel.

image

Ponkhy (approved):
LGTM

charliemoss:
Worked well. I don't know how to "review" so I just leave comments.

I had to add a BIND. Host Path: /path/to/photos  Mount Path: /photos, but you probably know that already.

PR #276, opened 2023-03-16
Add Dokuwiki to Easypanel 🎉 

Supernova3339:


Ponkhy:
Hey, maybe also add a user input for the `TZ` variable and have `Etc/UTC` as the default value? Since having the correct timezone for this is quite important, that way it's a little easier for the user.

Supernova3339:
> Hey, maybe also add a user input for the `TZ` variable and have `Etc/UTC` as the default value? Since having the correct timezone for this is quite important, that way it's a little easier for the user.

not implemented

Ponkhy (approved):
LGTM

PR #277, opened 2023-03-17
[Added]: Audiobookshelf

Ponkhy:


charliemoss:
The schema works perfectly. The only comment would be that maybe it should include "podcasts" as a separate BIND, as it is a separate Media Type (selectable when adding a library).

Supernova3339:
not everyone has podcasts as a BIND

Ponkhy:
> The schema works perfectly. The only comment would be that maybe it should include "podcasts" as a separate BIND, as it is a separate Media Type (selectable when adding a library).

Yes, I also could have gone with no default bind, but I wanted to give the user a example they can work with.

PR #284, opened 2023-03-18
add: Minecraft Server

berksmbl:


templates/minecraft-server/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+      title: App Service Name
+      default: minecraft-server
+    appServiceImage:
+      type: string
+      title: App Service Image
+      default: itzg/minecraft-server:2023.2.0
+    eula:
+      type: boolean
+      title: Please accept the Minecraft EULA
+      default: true
```
Ponkhy, line 37: Since it's a EULA, i'd recommend to have it on `false` by default and make it `required`.
```suggestion
      default: false
```

Ponkhy:
Maybe we could also pin the version with [`VERSION`](https://github.com/itzg/docker-minecraft-server#versions) and have another user input, so it's easier for people to start it with a specific version.

templates/minecraft-server/meta.yaml
berksmbl: If they do not accept this, the installation does not start.

Since I prepared it with the logic of 1 Click install, I want to have a ready-made template when the "Create" button is clicked directly without entering any value.

berksmbl: Even if I mark it as "required" because the boolean values are false and true, it returns a value if they don't accept it.

Ponkhy (approved):
LGTM

berksmbl:
Strange, it shouldn't fail.

Supernova3339 (approved):
LGTM

PR #290, opened 2023-03-20
[FIX] - Yet another Mattermost fix

Supernova3339:


Ponkhy:
I think it would also be good to add the domain property like below, since you have it in the user input and env.

```js
domains: [
  {
    name: input.domain,
  },
],
```

Ponkhy (approved):
LGTM

PR #291, opened 2023-03-20
[FIX] - Rocketchat fix 🚀 

Supernova3339:
Updates to v5.4.5 and fixes image. Deployment takes a minuteish to go green, then it will start up.


templates/rocketchat/assets/logo.jpg
deiucanta, line 1: the logo has a white background, the previous one had transparent background

Supernova3339: It's fake transparency, it's not actually transparent. Zoom in and you'll see

Supernova3339: that's the one I could find. feel free to change it when you merge

PR #294, opened 2023-03-20
Add Papercups to Easypanel 🎉 

Supernova3339:


Ponkhy (changes requested):
This needs to be added in ordner to make the template work, otherwise the database will not be filled with the required tables and relations.

```js
deploy: {
  command: "sh -c 'sleep 10 && /entrypoint.sh db createdb && /entrypoint.sh db migrate && echo 'running' && /entrypoint.sh run'",
},
```

Supernova3339:
> This needs to be added in ordner to make the template work, otherwise the database will not be filled with the required tables and relations.
> 
> ```js
> deploy: {
>   command: "sh -c 'sleep 10 && /entrypoint.sh db createdb && /entrypoint.sh db migrate && echo 'running' && /entrypoint.sh run'",
> },
> ```

I have that added on the deployment, forgot to add it to the template

Supernova3339:
Fixed

Ponkhy (approved):
LGTM

PR #295, opened 2023-03-21
Add Planka to Easypanel 🎉 

Supernova3339:


templates/planka/index.ts
Ponkhy, line 1: Wouldn't it be better to create these too?
[docker-compose.yml - Volumes](https://github.com/plankanban/planka/blob/master/docker-compose.yml#L16-L18)

Supernova3339: I'll add it in later

Supernova3339: added

PR #298, opened 2023-03-21
update: WordPress Redis Support

berksmbl:


Supernova3339:
Hi Berk! It might be better to build an image on epc's org for redis via `wordpress:6`.
You'll still have WordPress, but you can add your commands.

Supernova3339:
> Hi Berk! It might be better to build an image on epc's org for redis via `wordpress:6`. You'll still have WordPress, but you can add your commands.

Just send me a DM on disc and I'll send you an invite. -  no questions asked :)))))

Ponkhy:
Whenever possible we should use official images. I haven't tested the PR yet but I'd prefer to not build a custom one, at least for now.

berksmbl:
I deliberately avoided creating a new image.
It is desired to use official and stable images.

Supernova3339:
what you'll be doing is using the official image ontop of a custom
```
FROM wordpress:6

your code
```

templates/wordpress/index.ts
```diff
@@ -4,32 +4,59 @@ import { Input } from "./meta";
...
+
+  const fpm = input.appServiceImage.includes("fpm");
+  const appPort = fpm ? 9000 : 80;
+  const exec = fpm ? "php-fpm" : "apache2-foreground";
+  const alpine = input.appServiceImage.includes("alpine")
+    ? "apk add $PHPIZE_DEPS && "
+    : "";
+  const cmd =
+    alpine +
+    "pecl install redis && docker-php-ext-enable redis && docker-entrypoint.sh " +
```
deiucanta, line 24: This is a smart solution, but I don't want to have people install things on every service restart. Ideally, this should be a separate docker image or a github repo with a Dockerfile. I'm happy to host it on `https://github.com/esaypanel-io/wordpress-redis` if you want to build it.

berksmbl: Actually, a small change is enough [link](https://github.com/berksmbl/wordpress-redis)

Supernova3339:
told you to do a custom **Dockerfile**
seems like you did
:)

PR #299, opened 2023-03-22
Add Octobox to Easypanel 🎉 

Supernova3339:


templates/octobox/index.ts
```diff
@@ -0,0 +1,49 @@
...
+  const databasePassword = randomPassword();
+  const redisPassword = randomPassword();
+
+  services.push({
+    type: "app",
+    data: {
+      projectName: input.projectName,
+      serviceName: input.appServiceName,
+      env: [
+        `RAILS_ENV=production`,
```
Ponkhy, line 19: This either needs to be changed to `development` or the `SECRET_KEY_BASE` env needs to be added.

templates/octobox/index.ts
```diff
@@ -0,0 +1,49 @@
...
+        `GITHUB_CLIENT_ID=${input.githubClient}`,
+        `GITHUB_CLIENT_SECRET=${input.githubSecret}`,
+        `OCTOBOX_DATABASE_NAME=${input.projectName}`,
+        `OCTOBOX_DATABASE_USERNAME=postgres`,
+        `OCTOBOX_DATABASE_PASSWORD=${databasePassword}`,
+        `OCTOBOX_DATABASE_HOST=${input.projectName}_${input.databaseServiceName}`,
+        `REDIS_URL=PAPERLESS_REDIS=redis://default:${redisPassword}@${input.projectName}_${input.redisServiceName}:6379`,
```
Ponkhy, line 26: Redis Server deployment is missing.

Ponkhy (changes requested):
When it starts the Easypanel domain will be blocked. Maybe disable it or add a domain property and add it to the allow list?

Supernova3339:
Whoops I'll fix that sorry

Ponkhy:
The following env also need to be added `ALLOW_ALL_HOSTNAMES=true`.

deiucanta:
@Supernova3339 do you want to fix this?

Supernova3339:
> @Supernova3339 do you want to fix this?

Will be fixed ASAP

PR #301, opened 2023-03-22
Add Gladys Assistant to Easypanel 🎉 

Supernova3339:


Ponkhy:
This template does not start, could you please also test it?
My suspicion is it needs to run as `privileged` and it's currently not possible to run these as template.

Ahson-Shaikh:
This is still in draft, closing it and creating a separate PR for this.

PR #307, opened 2023-03-24
add: NodeBB

berksmbl:


Ponkhy (approved):
LGTM

templates/nodebb/assets/screenshot.png
Supernova3339, line 1: I would recommend you add additional screenshots.
Examples could include the front or backend.

Supernova3339: Let me know if you wish to implement this so I can approve it

berksmbl: The demo will suffice.

Supernova3339 (approved):
LGTM

Tc-001:
Looks like it is not possible to add plugins with the default configuration, as they get overwritten once the server restarts.
Adding a volume mount pointing to `/usr/src/app/node_modules` seems to make it work, but I am unsure if that is the "correct" solution.

berksmbl:
> Looks like it is not possible to add plugins with the default configuration, as they get overwritten once the server restarts. Adding a volume mount pointing to `/usr/src/app/node_modules` seems to make it work, but I am unsure if that is the "correct" solution.

I'm checking.

PR #309, opened 2023-03-25
[UPDATE] - Chatwoot to v2.15.0

Supernova3339:


Ponkhy (changes requested):
Since the domain will be defined in the user input, I think we should also add the following.
```js
domains: [
  {
    name: input.domain,
  },
],
```

Ponkhy (approved):
LGTM

deiucanta:
Already updated to 2.16

PR #314, opened 2023-03-26
add: Mastodon

berksmbl:


Ponkhy:
What does speak against using the official docker images in this case, because it requires more apps to set up or something different?

Supernova3339:
VAPID keys, that's my guess

berksmbl:
Yes, there are codes that need to be run only during the initial setup and I couldn't find a way to run the codes only once.

berksmbl:
> 

No, that's not the issue. I can solve it with this example.
```js
const crypto = require("crypto");
const ecdh = crypto.createECDH("prime256v1");
ecdh.generateKeys();
const vapidPublicKey = ecdh.getPublicKey("base64")
const vapidPrivateKey = ecdh.getPrivateKey("base64")
```

berksmbl:
> that's good as I was using a complex library easypanel wasn't using

"crypto" is a default library that comes with the system. The correct library to use for this process is the "web-push" library. Unfortunately, it needs to be added in order for me to be able to use it.

Supernova3339:
let andrei know! If it makes things easier, I'm sure he'd add it. I was

Supernova3339:
sorry, the end of that was an accidental copy-paste :)

charliemoss:
> Is there a default admin login? I asked ChatGPT, and it told me to create an account and then in the profile select moderator... etc.

I got it, run -
RAILS_ENV=production bin/tootctl accounts create \
  alice \
  --email alice@example.com \
  --confirmed \
  --role Owner

charliemoss:
What is the Streaming Domain? I am struggling to find that. Great job by the way.

berksmbl:
> Is there a default admin login? I asked ChatGPT, and it told me to create an account and then in the profile select moderator... etc.

Instructions for this are already available with the relevant explanation. [Link.](https://github.com/berksmbl/easypanel-templates/blob/add-mastodon/templates/mastodon/meta.yaml#L3)

Supernova3339:
Insidious has a hard-coded user and Andrei doesn't like me not using the
app service

I will add it myself later, no worries.
I would appreciate if you could try to get Clearflask running

templates/mastodon/index.ts
```diff
@@ -0,0 +1,126 @@
...
+  Services,
+} from "~templates-utils";
+import { Input } from "./meta";
+
+export function generate(input: Input): Output {
+  const services: Services = [];
+  const databasePassword = randomPassword();
+  const redisPassword = randomPassword();
+
+  const crypto = require("crypto");
```
deiucanta, line 14: This will not work in Easypanel as the `crypto` dependecy will not exist in the frontend env. Instead, you can update the form to require the public/private keys from the user. Also, add simple description on how to obtain a key pair.

berksmbl: Hello, you are already using this dependency in the "[randomPassword](https://github.com/easypanel-io/templates/blob/main/utils/schema.ts#L1)" function. Therefore, I believe there won't be any issues on the front-end.

deiucanta: You're right, I forgot about it.

PR #316, opened 2023-03-26
Add Chatpad to Easypanel 🎉 

Supernova3339:


templates/chatpad/index.ts
```diff
@@ -0,0 +1,31 @@
...
+        type: "image",
+        image: input.appServiceImage,
+      },
+      proxy: {
+        port: 80,
+        secure: true,
+      },
+      mounts: [
+        {
+          type: "volume",
```
deiucanta, line 22: no need for a mount as this will make updating the version impossible

templates/chatpad/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+      type: string
+      title: Project Name
+    appServiceName:
+      type: string
+      title: App Service Name
+      default: chatpad
+    appServiceImage:
+      type: string
+      title: App Service Image
+      default: ghcr.io/deiucanta/chatpad:b29ca06fb220ce74b5b5a051d414ec82fcdb2829
```
deiucanta, line 32: in this case we can use latest since new versions cannot break anything - there is no db connection, there are no mounts, no ports, so nothing to break

Ponkhy: Agreed, since the data is stored locally in the browsers storage.

Ponkhy (approved):
LGTM

PR #321, opened 2023-03-29
Add Cachet to Easypanel 🎉 

Supernova3339:
This adds Cachet to Easypanel
The image used is NOT official because the official image is broken ( shocker I know )
If this ends up unmaintained after a few years I might end up rewriting it as this is an awesome app!


Ponkhy (changes requested):
The following needs to be added to the env:
```js
DB_DATABASE=${input.projectName}
```

Supernova3339:
have you tried to deploy it? this is built with laravel 5

Ponkhy:
> have you tried to deploy it? this is built with laravel 5

Yes, but of course it doesn't work unless you add the env. from my message before

Supernova3339:
I will fix that

deiucanta:
@Supernova3339 is this fixed?

Supernova3339:
> @Supernova3339 is this fixed?

Fixed and working!

PR #329, opened 2023-04-04
Add Passbolt to Easypanel 🎉 

Supernova3339:
Another PR after a few days? Shocker!
But as always, you'll get the best of the best.
Will I end up making my own password manager? probably not but maybe in a few years.


templates/passbolt/index.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  services.push({
+    type: "app",
+    data: {
+      projectName: input.projectName,
+      serviceName: input.appServiceName,
+      env: [
+        `APP_FULL_BASE_URL=https://${input.domain}`,
+        `DATASOURCES_DEFAULT_HOST=${input.projectName}_${input.databaseServiceName}`,
+        `DATASOURCES_DEFAULT_USERNAME=postgres`,
```
Ponkhy, line 22:
```suggestion
        `DATASOURCES_DEFAULT_USERNAME=mysql`,
```

templates/passbolt/index.ts
```diff
@@ -0,0 +1,69 @@
...
+        `DATASOURCES_DEFAULT_DATABASE=${input.projectName}`,
+        `PASSBOLT_REGISTRATION_PUBLIC=false`,
+        `PASSBOLT_SSL_FORCE=true`,
+      ].join("\n"),
+      source: {
+        type: "image",
+        image: input.appServiceImage,
+      },
+      deploy: {
+        command: "sleep 10 && ./docker-entrypoint.sh"
```
Ponkhy, line 33:
```suggestion
        command: "sleep 10 && /docker-entrypoint.sh"
```

templates/passbolt/meta.yaml
```diff
@@ -0,0 +1,43 @@
+name: Passbolt
+description: The open source password manager for teams
+instructions: null
```
Ponkhy, line 3:
```suggestion
instructions: In order to create the first admin account, the following needs to be executed inside the passpolt container "su -m -c "bin/cake passbolt register_user -u YOUR@EMAIL.COM -f YOURNAME -l SURNAME -r admin" -s /bin/sh www-data". Please follow further instructions of the command output.
```

Ponkhy (changes requested):

templates/passbolt/index.ts
```diff
@@ -0,0 +1,69 @@
...
+  Output,
+  randomPassword,
+  randomString,
+  Services,
+} from "~templates-utils";
+import { Input } from "./meta";
+
+export function generate(input: Input): Output {
+  const services: Services = [];
+  const secret = randomString(512);
```
Ponkhy, line 11: Not used anywhere
```suggestion
```

templates/passbolt/meta.yaml
Ponkhy: I think this is quite important for the end user and should be implemented.

Ponkhy (approved):
LGTM

PR #345, opened 2023-04-12
Add Cusdis to Easypanel 🎉 

Supernova3339:


templates/cusdis/index.ts
```diff
@@ -0,0 +1,57 @@
...
+        `USERNAME=${input.username}`,
+        `PASSWORD=${input.password}`,
+        `NEXTAUTH_URL=https://${input.domain}`,
+        `DB_TYPE=pgsql`,
+        `DB_URL=postgres://postgres:${databasePassword}@${input.projectName}_${input.databaseServiceName}:5432/${input.projectName}?sslmode=disable`,
+        `JWT_SECRET=${secret}`,
+        `IS_HOSTED=true`,
+        `NODE_ENV=production`,
+        `HOST=https://${input.domain}`,
+        `SMTP_SENDER=Cusdis Notification<notification@${input.domain}>`,
```
Ponkhy, line 29:
```suggestion
```

I don't think this is needed, unless you also add every other SMTP property to enable the notifications.

https://github.com/djyde/cusdis/blob/1fb82d5673dacee7961b87ee543d9a671c645e8e/public/doc/features/notification.md?plain=1#L25

Supernova3339:
this is set to the cloud domain if it is not configured, and as DO supports
this it is needed to ensure proper sending and prevent bad reputation on
the IP sending emails

Ponkhy:
> this is set to the cloud domain if it is not configured, and as DO supports this it is needed to ensure proper sending and prevent bad reputation on the IP sending emails

How does it send emails without a smtp server or similar configured, or do they have some kind of internal mailer?

ivanciucdragos1 (approved):

PR #356, opened 2023-05-09
Renameregistry

ivanciucdragos1:


templates/registry/meta.yaml
```diff
@@ -37,7 +37,7 @@ schema:
     appServiceName:
       type: string
       title: App Service Name
-      default: registry
+      default: docker registry
```
deiucanta, line 40: the name cannot have a "space", it should be `docker-registry` or simply `registry`

templates/registry/meta.yaml
```diff
@@ -1,4 +1,4 @@
-name: Registry
+name: Docker registry
```
deiucanta, line 1: use `Docker Registry` (capitalized)

templates/registry/meta.yaml
deiucanta, line 1: rename the folder as well

PR #383, opened 2023-06-26
LangFlow and Flowise

serban-alexandru:


viniciusvinicius:
Hi Serban, I am using Flowise on Easypanel, first of all thanks for the implementation. 
Do you know if it is possible to update the Flowise version to 2.1.1? (my version shows 1.8.0). 
If yes, can you explain me how to do it? Thanks!

PR #385, opened 2023-06-30
qdrant template

serban-alexandru:


templates/qdrant/index.ts
```diff
@@ -0,0 +1,29 @@
...
+      projectName: input.projectName,
+      serviceName: input.appServiceName,
+      source: {
+        type: "image",
+        image: input.appServiceImage,
+      },
+      deploy: {
+        replicas: 1,
+        command: null,
+        zeroDowntime: true,
```
deiucanta, line 19: these are default values, not required

templates/qdrant/meta.yaml
```diff
@@ -0,0 +1,33 @@
...
+    projectName:
+      type: string
+      title: Project Name
+    appServiceName:
+      type: string
+      title: App Service Name
+    appServiceImage:
+      type: string
+      title: App Service Image
+      default: qdrant/qdrant
```
deiucanta, line 33: use fixed version, not latest

templates/qdrant/index.ts
```diff
@@ -0,0 +1,24 @@
...
+    type: "app",
+    data: {
+      projectName: input.projectName,
+      serviceName: input.appServiceName,
+      source: {
+        type: "image",
+        image: input.appServiceImage,
+      },
+      proxy: {
+        port: 80,
```
deiucanta, line 17: wrong port number, it should be 6333

templates/qdrant/index.ts
```diff
@@ -0,0 +1,24 @@
...
+        secure: true,
+      },
+    },
```
deiucanta, line 23: volume mount is missing

PR #389, opened 2023-07-02
Update Ghost v5.53.3

siriokun:


templates/ghost/index.ts
```diff
@@ -37,7 +37,7 @@ export function generate(input: Input): Output {
   });
 
   services.push({
-    type: "mariadb",
```
deiucanta, line 40: why did you switch from mariadb?

templates/ghost/index.ts
```diff
@@ -18,7 +18,7 @@ export function generate(input: Input): Output {
         `url=https://${input.domain}`,
         `database__client=mysql`,
         `database__connection__host=${input.projectName}_${input.databaseServiceName}`,
-        `database__connection__user=mariadb`,
```
deiucanta, line 21: The db user is not the project name. Can you make sure this works?

templates/ghost/meta.yaml
```diff
@@ -43,8 +48,12 @@ schema:
...
-      default: ghost:5.35.1-alpine
+      default: ghost:5.53.3-alpine
     databaseServiceName:
       type: string
       title: Database Service Name
       default: ghost-db
+    databaseType:
+      type: string
+      title: Database Type
+      default: mysql
```
deiucanta, line 59: Do you want to support multiple dbs? Otherwise, this field is not useful.

templates/ghost/index.ts
siriokun: Start from Ghost 5.0, MariaDB is not an officially supported database for Ghost.

https://github.com/TryGhost/Ghost/issues/14446

templates/ghost/meta.yaml
siriokun: Ghost 5.0 only support MySQL 8 on Production,

https://ghost.org/docs/faq/supported-databases/

This option is indeed not usefull. I will remove it now.

siriokun:
Added MariaDB support from @berksmbl for now, need to consider that starting from Ghost 5.0, MariaDB is not supported on Production https://ghost.org/docs/faq/supported-databases/

berksmbl:
> Added MariaDB support from @berksmbl for now, need to consider that starting from Ghost 5.0, MariaDB is not supported on Production https://ghost.org/docs/faq/supported-databases/

@siriokun I did not know that Ghost had a usage suggestion in this direction. In this case, it would be more logical to use only MySQL.

siriokun:
Hi @deiucanta we can use MySQL for now. Can you review this?

PR #424, opened 2023-07-17
update: Botpress

berksmbl:
By default, duckling is running on port 8000, but the port value of the primary domain in the proxy via Easypanel is transmitted to the application via the PORT variable. duckling starts working on this port for this reason, which causes the port to fill up and the application to start on the upper port. It also causes the application to not be able to access the duckling.


Ahson-Shaikh:
We really don't support having latest tags. And the image is already updated.

PR #437, opened 2023-08-02
Fix Chatwoot Worker

meaziz:
Chatwoot doesn't start the SideKiq Worker which manages the realtime communication between web chats.

Added it to the deploy command


Supernova3339 (approved):
LGTM

matbrgz (approved):

CarefulGuru (changes requested):
That change doesn't work unfortunately. `bundle exec rails s -p 3000 -b 0.0.0.0` is a never-ending task, so chaining `bundle exec sidekiq -C config/sidekiq.yml` in sequentially makes no difference. The `bundle exec rails` task never ends, so it never actually starts the sidekiq server. Changing this to `bundle exec rails db:chatwoot_prepare && (bundle exec rails s -p 3000 -b 0.0.0.0 &) && bundle exec sidekiq -C config/sidekiq.yml` solves that issue. This allow sidekiq and rails to run at the same time.

briannezhad:
When this real-time communication will be resolved with this merge?

CarefulGuru:
@briannezhad this will solve that issue when it is merged. However, for the time being, simply change the deploy command to the one in my comment above. Go to the Advanced tab in your chatwoot app:
image
Then, paste this command: `bundle exec rails db:chatwoot_prepare && (bundle exec rails s -p 3000 -b 0.0.0.0 &) && bundle exec sidekiq -C config/sidekiq.yml` into the "command" box 
image
Hit save, then hit the "deploy" button for the app. Should fix the issue!

briannezhad:
Works! Thank you @CarefulGuru

michaelvips:
@CarefulGuru @briannezhad for me this not works.

PR #476, opened 2024-01-03
updated & tested

serban-alexandru:


templates/minio/index.ts
```diff
@@ -0,0 +1,46 @@
...
+      mounts: [
+        {
+          type: "volume",
+          name: "data",
+          mountPath: "/data",
+        },
+      ],
+      domains: [
+        {
+          host: "$(EASYPANEL_DOMAIN)",
```
deiucanta, line 30:
```
host: "console.$(EASYPANEL_DOMAIN)",
```

templates/n8n/index.ts
```diff
@@ -9,6 +9,7 @@ export function generate(input: Input): Output {
     data: {
       projectName: input.projectName,
       serviceName: input.appServiceName,
+      env: ["WEBHOOK_URL=$(EASYPANEL_DOMAIN)"].join("\n"),
```
deiucanta, line 12: should we add https before the domain?

PR #499, opened 2024-03-03
update: Answer

berksmbl:
Updated application version to 1.2.0 and added automatic installation support


berksmbl:
To be opened collectively

PR #501, opened 2024-03-10
Add plane template

marktopper:
This templates allows to setup **[Plane](https://plane.so) v0.16-dev**.

# Why customize

This template uses file mounts to overwrite some logic in the proxy service.

- `/etc/nginx/nginx.conf.custom`: This is because the default nginx configuration in `makeplane/proxy` uses other internal host names (such as `api`, `web`, `space` and `minio`) rather than the full hostnames supported by EasyPanel (such as `PROJECTNAME_api`, `PROJECTNAME_web`, `PROJECTNAME_space` and `PROJECTNAME_minio`).
- `/custom-docker-entrypoint.sh`: This is used to apply the custom nginx configuration above.

# Template Issues

We are using the `$(EASYPANEL_DOMAIN)` option to automatically generate the domain names for the `proxy` and `minio` services. However, since this variable includes the service name, it applies the wrong domain names below:

1. The `proxy` service domain name, should ideally be `PROJECTNAME.INSTANCE.easypanel.host`, and not `PROJECTNAME-proxy.INSTANCE.easypanel.host`.
2. The `WEB_URL` environment in `api`, `beat-worker` and `worker` service should match the above domain name. Right now it applies the service name (`api`, `beat-worker` and `worker`) to it.
3. The domain names for the `minio` service, should ideally be something like one of the following options:
- `PROJECTNAME-minio.INSTANCE.easypanel.io + PROJECTNAME-minio-console.INSTANCE.easypanel.io`
- `PROJECTNAME-minio.INSTANCE.easypanel.io + console.PROJECTNAME-minio.INSTANCE.easypanel.io`
4. The `AWS_S3_ENDPOINT_URL` environment in `api`, `beat-worker` and `worker` service should match the above domain name. Right now it applies the service name (`api`, `beat-worker` and `worker`) to it.

# Configuration

Ideally we should add some more configuration options, such as:
- Wether or not `minio` service should be generated: Could be that the user prefers to use another S3 supported service, or already have a minio service running.
- Sentry DNS / Sentry Environtment

# Optimizations

1. Currently, I was not able to get the `api`, `beat-worker` and `worker` able to connect to the `minio` service using the internal hostname, therefor a domain for the `minio` service API is required, while it would be safer to use a internal endpoint for this.
2. Ideally, it would be nice if `makeplane/proxy` supported setting the hostnames by environment variables, then we can remove the custom nginx configuration and the custom entrypoint.

# Help

If anyone here is willing to make a PR for the domain name issues, configuration options or any other improvements, feel free. This works for my own usage, so I will not spend more time optimizing this template.


marktopper:
@deiucanta This has been updated to use the new `EASYPANEL_HOST` variable + the addition of more configuration options.

PR #503, opened 2024-03-17
add: 2FAuth

berksmbl:
2FAuth is a web based self-hosted alternative to One Time Passcode (OTP) generators like Google Authenticator, designed for both mobile and desktop.


serban-alexandru:
@deiucanta Ready for merge

PR #504, opened 2024-03-18
Add Cronicle Template

marktopper:
https://cronicle.net/


berksmbl (approved):
LGTM
I think that only the "path" and "https" fields in the domains field are unnecessary, Default already comes with these definitions. But this will not affect the work

marktopper:
> LGTM
> 
> > _I think that only the "path" and "https" fields in the domains field are unnecessary, Default already comes with these definitions. But this will not affect the work_

These has been removed now (commit [9fbc454](https://github.com/easypanel-io/templates/pull/504/commits/9fbc45430e6c0112f05f8cf7098b96baa3dd2169))

serban-alexandru:
@deiucanta Ready for merge

PR #511, opened 2024-03-19
Remove incorrect instructions

marktopper:
Instructions are wrong, after deploying, user will enter a setup stage on the Odoo site, from there they set the username and password.


berksmbl (changes requested):
Yeah, that seems about right. Can you upgrade to version `17.0` while you're at it?

marktopper:
Duplicate of #481

PR #526, opened 2024-05-09
Added NEXTAUTH_URL variable

alanfrigo:
Added the missing NEXTAUTH_URL variable at the Typebot Viewer App


Supernova3339 (changes requested):
Add your name to meta.yaml so we can credit you for the modification

Supernova3339 (approved):
LGTM

PR #565, opened 2024-10-07
Add jenkins template

MuhammadAhsanDonuts:


templates/jenkins/meta.yaml
```diff
@@ -0,0 +1,85 @@
...
+schema:
+  type: object
+  required:
+    - appServiceName
+    - appServiceImage
+  properties:
+    appServiceName:
+      type: string
+      title: App Service Name
+      default: jenkins_web
```
deiucanta, line 29: use `jenkins` instead of `jenkins_web` - when we have only one service it doesn't make sense to add suffixes

PR #585, opened 2024-10-29
Rabbitmq Suggested Fix

Ahson-Shaikh:
After an hour of contemplation, I decided to build it like this. The previous logic was broken. We can proceed with either this logic or someone can suggest any changes. Thanks.

### DIFF
--- a/templates/rabbitmq/index.ts
+++ b/templates/rabbitmq/index.ts
@@ -29,16 +29,27 @@ export function generate(input: Input): Output {
     });
   }
 
+  let image = "rabbitmq:3.13.7";
+
+  if (input.appServiceImage) {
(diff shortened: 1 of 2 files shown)


deiucanta:
@Ahson-Shaikh please explain what is problem we're trying to fix here

Ahson-Shaikh:
@deiucanta There are two different set of images for different versions of rabbitmq. One is for the CLI and the other is for the Management. The checkbox enables the management UI for rabbitmq, but the current logic only changes it to the management image if there is no default value inside the appServiceImage and UI box is checked - I came up with this and a set of instructions. Let me know if we can improve it.

Ahson-Shaikh:
I also believe that creating two separate RabbitMQs, one for the CLI based version and other for the management UI version to make it easier and quicker. That would be better.

PR #587, opened 2024-10-31
Chatwoot Sideqik container access to volumes fixed. 

Ahson-Shaikh:
Chatwoot uses a sideqik container, that was having issues while accessing the data produced within the chatwoot volumes app and data directory. Fixed now.


michaelvips (approved):
Works for me

Ahson-Shaikh:
Hi Michael - Thanks for pointing that. I was trying to keep the original volumes defined in the template intact while providing the path to the sidequik. 

image

But, I have double-checked, it is not necessary for sidequik. I'll change it right now.

PR #605, opened 2024-11-16
Template Added Qbittorent

Ahson-Shaikh:
"qBittorrent is a powerful, open-source BitTorrent client designed to provide
  an efficient and feature-rich experience for downloading and sharing files. It
  offers a user-friendly interface, robust functionality, and cross-platform
  compatibility. qBittorrent supports advanced features like sequential
  downloading, IP filtering, RSS feed support, and a built-in search engine for
  finding torrents. It also includes a web-based interface for managing
  downloads remotely. Designed to balance performance, simplicity, and advanced
  capabilities, qBittorrent is ideal for users seeking a secure and reliable
  torrenting solution."


templates/qbitorrent/index.ts
```diff
@@ -0,0 +1,49 @@
...
+        },
+        {
+          type: "volume",
+          name: "config",
+          mountPath: "/config",
+        },
+      ],
+      ports: [
+        {
+          published: 6881,
```
deiucanta, line 35: I think you should make these ports configurable using the input fields.

templates/qbitorrent/index.ts
```diff
@@ -0,0 +1,49 @@
...
+        {
+          type: "volume",
+          name: "config",
+          mountPath: "/config",
+        },
+      ],
+      ports: [
+        {
+          published: Number(input.qBittorrentPort),
+          target: Number(input.qBittorrentPort),
```
deiucanta, line 36: the target port should be hardcoded; if the user changes it, it will not work; only the published port should be configurable

Ahson-Shaikh: @deiucanta I have hardcoded the targeted ports. Thanks!

PR #608, opened 2024-11-17
Fixed Pocketbase Access Issue. 

Ahson-Shaikh:
Fixed pocket base access issue dashboard and api, updated image to a static version.


templates/pocketbase/index.ts
```diff
@@ -15,6 +15,12 @@ export function generate(input: Input): Output {
       domains: [
         {
           host: "$(EASYPANEL_DOMAIN)",
+          path: "/_/",
```
deiucanta, line 18: are you sure this is the correct path?

PR #617, opened 2024-11-27
HackMD template

serban-alexandru:


Ahson-Shaikh:
@serban-alexandru The template requires couple of more envs for login. 
CMD_ALLOW_EMAIL_REGISTER=true
CMD_EMAIL=true

Otherwise there is a default SAML auth, that needs to be configured. 
https://hackmd.io/s/codimd-configuration

I was unable to login before adding these envs.

serban-alexandru:
@Ahson-Shaikh I've added the variables to the env

PR #620, opened 2024-11-27
Ombi template

serban-alexandru:


Ahson-Shaikh:
@serban-alexandru The template looks perfect! Just the latest tag should be changed.

serban-alexandru:
@Ahson-Shaikh I've adjusted the version

PR #627, opened 2024-11-28
Raneto template

serban-alexandru:


Ahson-Shaikh:
@serban-alexandru Template works good! Only the latest tag needs to be changed.

PR #628, opened 2024-11-29
Template Added Private Bin

Ahson-Shaikh:
PrivateBin is a minimalist, open-source, self-hosted pastebin solution
  designed for secure and private data sharing. It features end-to-end
  encryption, ensuring that your shared content remains confidential. With its
  simple interface and focus on privacy, PrivateBin is an excellent tool for
  individuals and teams who value secure communication.


templates/privatebin/index.ts
```diff
@@ -0,0 +1,47 @@
...
+      ],
+      mounts: [
+        {
+          type: "volume",
+          name: "data",
+          mountPath: "/srv/data",
+        },
+        {
+          type: "file",
+          content:
```
deiucanta, line 38: move this to a multi line string in a const

Ahson-Shaikh: @deiucanta I have fixed that and tested.

PR #629, opened 2024-11-30
Update meta.yaml

tempusthales:
The current version of Photo Prism in Easy Panel is outdated by two versions.  Current release version is 26.0.6


Ahson-Shaikh:
@tempusthales We don't prefer using latest tags as they can go incompatible with the static configurations. Can you please change it with a static tag?

PR #674, opened 2024-12-21
template fixed pocketbase

Ahson-Shaikh:


templates/pocketbase/index.ts
```diff
@@ -15,6 +15,12 @@ export function generate(input: Input): Output {
       domains: [
         {
           host: "$(EASYPANEL_DOMAIN)",
+          path: "/_/",
```
deiucanta, line 18: I don't think this change is needed.

deiucanta: By default all traffic is proxied to the service.

With this change you try to proxy only `/_` and `/api`. You use the same service and the same port. It doesn't change the default behavior - it simply limits the paths that it proxies.

deiucanta: @Ahson-Shaikh please confirm and fix this

Ahson-Shaikh: image
image

I have tested this multiple times and the change I have recommended works. Otherwise, the neither the app can access the api nor I can access the app on browser.

deiucanta: @Ahson-Shaikh are you clear on this now? please close the PR

PR #687, opened 2024-12-26
Template Daapd Added

Ahson-Shaikh:
Daapd is a robust, open-source DAAP (Digital Audio Access Protocol) server
  that allows users to share music libraries over a network. It is designed to
  work seamlessly with iTunes, Roku SoundBridges, and other DAAP clients,
  providing an efficient and reliable way to stream audio content across
  devices. Daapd also supports streaming internet radio, podcasts, and AirPlay
  devices. With its powerful backend, Daapd ensures smooth performance, a wide
  range of supported formats, and advanced features like library management and
  transcoding.


templates/daapd/meta.yaml
```diff
@@ -0,0 +1,70 @@
...
+    - appServiceImage
+  properties:
+    appServiceName:
+      type: string
+      title: App Service Name
+      default: daapd
+    appServiceImage:
+      type: string
+      title: App Service Image
+      default: lscr.io/linuxserver/daapd:28.10.20241218
```
deiucanta, line 38: sometimes you use lscr.io and sometimes you omit it

for linuxserver images you can always use it, it's a proxy/mirror and helps users avoid docker hub rate limits

PR #736, opened 2025-01-16
Added Template Picoshare

Ahson-Shaikh:
Picoshare is a lightweight, self-hosted file-sharing platform designed for
  simplicity and privacy. It provides a minimalistic interface for uploading,
  sharing, and managing files, without the complexity of traditional cloud
  storage services. Picoshare emphasizes ease of use and security, making it an
  ideal solution for personal or small team file-sharing needs. It supports a
  variety of file types and provides a secure, private environment for sharing
  files with full control over your data.


templates/picoshare/assets/screenshot.png
deiucanta, line 1: bad aspect ratio

templates/picoshare/meta.yaml
```diff
@@ -0,0 +1,79 @@
...
+    - authPassphrase
+  properties:
+    appServiceName:
+      type: string
+      title: App Service Name
+      default: picoshare
+    appServiceImage:
+      type: string
+      title: App Service Image
+      default: mtlynch/picoshare
```
deiucanta, line 38: use static version, since it has a data volume

Ahson-Shaikh: @deiucanta fixed both issues.

PR #776, opened 2025-02-04
Added Template RSS-Bridge

Ahson-Shaikh:
RSS-Bridge is a self-hosted, open-source tool that generates RSS and Atom
  feeds for websites that do not provide them natively. It enables users to
  extract and aggregate content from various web pages, social media platforms,
  and other online sources into a single, unified RSS feed. This allows users to
  stay updated with the latest content from multiple sources without manually
  visiting each site. RSS-Bridge supports various website connectors, filtering
  options, and output formats to provide a customized feed experience.


templates/rss-bridge/meta.yaml
```diff
@@ -0,0 +1,81 @@
...
+    - appServiceImage
+  properties:
+    appServiceName:
+      type: string
+      title: App Service Name
+      default: rss-bridge
+    appServiceImage:
+      type: string
+      title: App Service Image
+      default: rssbridge/rss-bridge:latest
```
deiucanta, line 35: don't use latest

PR #790, opened 2025-02-12
Added Template Carbone

Ahson-Shaikh:
Carbone is a high-performance, self-hosted document generation engine. It
  allows you to generate reports, invoices, and documents in various formats
  (e.g., PDF, DOCX, XLSX) using JSON data and template-based rendering. Carbone
  is designed for scalability and speed, making it ideal for automation, bulk
  document generation, and seamless integration into various systems and
  workflows. It supports custom templates and advanced features like charts,
  sub-reports, and conditional formatting to meet a wide range of business
  needs.


templates/carbone/meta.yaml
```diff
@@ -0,0 +1,80 @@
...
+      type: string
+      title: App Service Name
+      default: carbone
+    appServiceImage:
+      type: string
+      title: App Service Image
+      default: carbone/carbone-ee:full-4.25.5
+    carboneKey:
+      type: string
+      title: Carbone License Key
```
deiucanta, line 40: can this be optional? most people will want to try it quickly and if they like it they might buy a license

Ahson-Shaikh: @deiucanta Fixed

PR #791, opened 2025-02-12
Added Template Teable

Ahson-Shaikh:
Teable is a powerful, self-hosted service designed for managing team
  collaboration and projects efficiently. It offers a clean and user-friendly
  interface, supports plugins for extended functionality, and provides detailed
  analytics on team activities. With Teable, you can maintain full control over
  your data and workflows.


templates/teable/meta.yaml
```diff
@@ -0,0 +1,83 @@
...
+    - migrationServiceImage
+  properties:
+    appServiceName:
+      type: string
+      title: App Service Name
+      default: teable
+    appServiceImage:
+      type: string
+      title: App Service Image
+      default: ghcr.io/teableio/teable:latest
```
deiucanta, line 41: Let's not use latest

Also, you should add a checkbox for arm64. If checked, add a `-arm64` at the end of the image.

PR #816, opened 2025-02-27
Added Template Zabbix

Ahson-Shaikh:
Zabbix is a powerful, self-hosted monitoring solution designed for tracking
  and analyzing system performance, network utilization, and application health.
  It provides real-time monitoring, customizable alerts, and detailed
  visualization tools to help administrators maintain system reliability. Zabbix
  supports agent-based and agentless monitoring, automation capabilities, and
  integration with third-party tools. With its flexible architecture, Zabbix is
  suitable for small to enterprise-level environments. The application includes
  a web-based dashboard and supports a variety of database backends, making it a
  robust solution for IT infrastructure monitoring.


templates/zabbix/meta.yaml
```diff
@@ -0,0 +1,82 @@
...
+      title: App Service Image
+      default: zabbix/zabbix-web-nginx-pgsql:alpine-trunk
+    zabbixServerServiceImage:
+      type: string
+      title: Zabbix Server Service Image
+      default: zabbix/zabbix-server-pgsql:alpine-trunk
+    zabbixAgentServiceImage:
+      type: string
+      title: Zabbix Agent Service Image
+      default: zabbix/zabbix-agent:alpine-trunk
```
deiucanta, line 47: this is similar to latest

please use a static version

Ahson-Shaikh: I have been trying to use the SHA values as the static tags, but EP keeps giving as no such image, but the deployment is successful and the containers ain't starting up. Not sure of the reason.

deiucanta: Find the reason, and open a bug report if needed or maybe it's not a bug. Those are not "SHA values", I think they simply used the SHA string for the tag name – I don't think there is anything special about it.

Ahson-Shaikh: @deiucanta I went on their GitHub Repo, I don't find the issue tab. Additionally, I signed up for their forums, but couldn't find a way to post - Seems like unauthorised.

deiucanta: No need to contact them. Debug the docker problem.

PR #822, opened 2025-03-05
Added Template Kokoro

Ahson-Shaikh:
Kokoro is a powerful, browser-based AI voice generator that lets you create
  natural-sounding voices without installing anything. Use it directly in your
  browser or self-host it for your own applications with OpenAI API
  compatibility!


Ahson-Shaikh:
Template rechecked!

PR #848, opened 2025-03-21
Added template lldap

Ahson-Shaikh:
LLDAP (Lightweight LDAP) is a simple, lightweight, high-performance LDAP
  server designed for authentication in small to medium-sized deployments. It
  provides an intuitive web UI for user management and a complete implementation
  of LDAP search and bind operations. LLDAP is perfect for self-hosted
  applications that require LDAP authentication without the complexity of
  full-featured LDAP servers like OpenLDAP. It stores user data in an embedded
  SQLite database by default but also supports PostgreSQL, making it easy to set
  up and maintain. With features like password hashing, account
  creation/deletion, group management, and MFA support, LLDAP offers a modern
  solution for identity management in your self-hosted infrastructure.


Ahson-Shaikh:
Template Checked!

PR #851, opened 2025-03-24
Updated Poste Links

Ahson-Shaikh:
Fixed Poste Links

### DIFF
--- a/templates/poste/meta.yaml
+++ b/templates/poste/meta.yaml
@@ -11,12 +11,12 @@ changeLog:
     description: first release
 links:
   - label: Documentation
-    url: https://poste.io/docs
+    url: https://poste.io/doc/
   - label: Github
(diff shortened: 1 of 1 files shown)


templates/poste/meta.yaml
```diff
@@ -11,12 +11,12 @@ changeLog:
     description: first release
 links:
   - label: Documentation
-    url: https://poste.io/docs
+    url: https://poste.io/doc/
   - label: Github
-    url: https://github.com/analogic/poste.io
+    url: https://github.com/dirtsimple/poste.io
```
deiucanta, line 16: this is not the official repo

Ahson-Shaikh: There is no official github I could find. I can skip that if you suggest. @deiucanta

Ahson-Shaikh:
It is no longer being maintained.

PR #875, opened 2025-04-09
Added Template LivestreamDVR

Ahson-Shaikh:
LiveStreamDVR (formerly TwitchAutomator) is a powerful tool for automatically recording and archiving live streams from Twitch. It provides a web interface to manage recordings, schedule streams to record, and organize your VOD library.


Ahson-Shaikh:
The official repo doesn't have the static tag.

PR #888, opened 2025-04-11
Added Template LibreNMS

Ahson-Shaikh:
LibreNMS is a fully featured network monitoring system that provides a wealth of features and device support.
  It includes support for multiple authentication methods, API access, alerting, and automatic discovery.


Ahson-Shaikh:
@deiucanta This is not production ready, I'm along with the client gonna check this if good to go.

PR #916, opened 2025-04-25
Fixed Plausible Template SMTP

Ahson-Shaikh:
Fixed SMTP Plausible.


templates/plausible/meta.yaml
```diff
@@ -36,6 +36,11 @@ schema:
     - appServiceImage
     - databaseServiceName
     - clickhouseServiceName
+    - mailerEmail
```
deiucanta, line 39: should these be required? someone could use/test plausible without configuring smtp

PR #928, opened 2025-05-02
Added Template Zusam

Ahson-Shaikh:
Zusam is a free and open-source way to self-host private forums for friends and family. It's a simple, modern, and privacy-focused platform for group discussions.


serban-alexandru:
@deiucanta @Ahson-Shaikh LGTM

PR #935, opened 2025-05-05
Added Template SteamCMD

Ahson-Shaikh:
SteamCMD is a command-line version of the Steam client for installing and updating Steam games and dedicated servers. It's perfect for managing game servers and game installations.


serban-alexandru:
@Ahson-Shaikh Same here. I see that we don't have db/volume but what if it will have someday?

serban-alexandru:
@deiucanta LGTM. I'm not sure about the version but being a CMD I think it should work fine.

PR #936, opened 2025-05-05
Added Template WatchTower

Ahson-Shaikh:
Watchtower is a process for automating Docker container base image updates. It monitors running containers and automatically updates them when their base image changes.


serban-alexandru:
@Ahson-Shaikh 
- has latest image tag - I think it should be 1.7.1
- https://containrrr.dev/watchtower - website - add in meta.yaml

Ahson-Shaikh:
@serban-alexandru Fixed site, but tag can be latest since no volume or db is involved.

serban-alexandru:
@Ahson-Shaikh Makes sense, but what if in a future version a db/volume will be required and it won't work properly?

PR #939, opened 2025-05-07
Added Template DoubleCommander

Ahson-Shaikh:
A cross-platform open-source file manager with two panels side by side. It is inspired by Total Commander and features a modern interface with extensive file management capabilities.


serban-alexandru:
@Ahson-Shaikh please add website in meta.yaml (https://doublecmd.sourceforge.io)

PR #943, opened 2025-05-08
Added Template AlertIO

Ahson-Shaikh:
Alerta is a monitoring system that consolidates and deduplicates alerts from multiple monitoring systems.


serban-alexandru:
@Ahson-Shaikh LGTM

PR #944, opened 2025-05-09
Added Template Adguard Home Sync

Ahson-Shaikh:
AdGuard Home Sync is a tool designed to synchronize filters, blocklists, and configuration between multiple AdGuard Home instances. It allows you to maintain consistent filtering across different AdGuard Home deployments by synchronizing configuration from an origin instance to one or more replica instances.


serban-alexandru:
@Ahson-Shaikh 
LGTM. maybe a full size screenshot, but I think it's ok

PR #967, opened 2025-05-29
Added template openpanel

Ahson-Shaikh:
OpenPanel is a powerful open-source product analytics platform that helps you
  track and analyze user behavior. Built with modern technologies, it provides
  real-time insights, custom event tracking, and intuitive dashboards. Features
  include user segmentation, funnel analysis, retention metrics, and
  comprehensive reporting capabilities. Self-hosted solution that gives you full
  control over your data while maintaining privacy and security.


templates/openpanel/index.ts
```diff
@@ -0,0 +1,205 @@
...
+        reverse_proxy {$API_URL}:3000
+    }
+
+    reverse_proxy /* {$DASHBOARD_URL}:3000
+}
+
+http://worker.{$PRIMARY_DOMAIN} {
+
+    basic_auth {
+        admin $2b$10$wZt0R0wkz0QaKe0HdLd2mO8gnqyz8QEt9bSzc7apFrfWHdWI.I0ZG
```
deiucanta, line 50: this seems to be hardcoded

templates/openpanel/index.ts
```diff
@@ -0,0 +1,205 @@
...
+        reverse_proxy {$API_URL}:3000
+    }
+
+    reverse_proxy /* {$DASHBOARD_URL}:3000
+}
+
+http://worker.{$PRIMARY_DOMAIN} {
+
+    basic_auth {
+        admin ${basicAuthHash}
```
deiucanta, line 50: a random hash will not work, basic auth hash is based on the password you want to set

also, we have this built in our services, so no need to add it in caddy

also, we don't need caddy, just use domains for each service

PR #1013, opened 2025-07-03
Fixed the Alltube Version Error

Ahson-Shaikh:
The Alltube Application has been archived by the maintainer. There were errors with the Python Library being using inside the container and the version update was also necessary. I couldn't find any updated one, so fixed the error in the code and rebuild the image.


Ahson-Shaikh:
The project has been archieved.

PR #1019, opened 2025-07-09
Updated Evolution API

Ahson-Shaikh:
Updated to the correct image and the latest updated tag.


moskoweb (approved):
Change for "latest"

moskoweb:
> There is another PR with an additional change, so closing this one. Moreover, @moskoweb we can not use latest tags due instability.

And tags "stable"?

PR #1066, opened 2025-08-19
Updated Twenty CRM Version to v1.3.0

Ahson-Shaikh:
Previously, there were errors on the migrations run for Twenty CRM. Opened up a ticket, and seems like the problem is not occurring and template is working as expected.


deiucanta (approved):
Approved.

PR #1102, opened 2025-09-19
Added Template Github Actions Runner

leogomide:


Ahson-Shaikh:
Avoid committing package.json or list.json. Additionally, please look at the other templates for the structure you would need to follow for this. Thanks!

PR #1154, opened 2025-10-24
A small QOL change that makes templates searchable

Supernova3339:
Makes templates searchable as a quality of life improvement for testing during development.


deiucanta:
Thanks @Supernova3339

PR #1218, opened 2025-12-16
Update MinIO service image default value from Bitnami to Minio

jefflombard:
Bitnami is no longer publishing on `bitnami/minio` moved to `minio/minio`


jefflombard:
Closing because it requires additional config. I believe port numbers need to get set straight and maybe some otherthings.

It runs fine, just fails on uploading

jefflombard:
Confirmed ports were mismatched. Need to update to `9000` and `9001` respectively

PR #1268, opened 2026-02-12
Updated & Fixed Template Faraday

Ahson-Shaikh:
- Switch from managed service for redis to the app service. 
- Faraday does have no way in their code where they may pass the password. Additionally, the full redis string is not operational either with it.


deiucanta (approved):
Reviewed against the template checklist.

PR #1303, opened 2026-03-19
Added Template GoClaw

moskoweb:
Multi-agent AI gateway with teams, delegation & orchestration. Single Go binary, 11+ LLM providers, 5 channels.


templates/goclaw/index.ts
```diff
@@ -0,0 +1,167 @@
...
+  const encryptionKey = input.encryptionKey || randomString(32);
+
+  // ── 1. PostgreSQL with pgvector ──
+  services.push({
+    type: "postgres",
+    data: {
+      serviceName: `${input.appServiceName}-db`,
+      password: databasePassword,
+      image: "pgvector/pgvector:pg18",
+      env: "PGDATA=/var/lib/postgresql/data",
```
Ahson-Shaikh: Why did you need to specify this? The managed postgres instance would have its mounts already defined

Ahson-Shaikh (commented):
Have you also tested the PR on Easypanel?

templates/goclaw/index.ts
moskoweb: It was necessary because PGVector 18 changed the file folder, and Easypanel didn't from version 16 to 18, and without that, it gave an error about the data loading folder.

PR #1304, opened 2026-03-19
Update and Improved Template OpenClaw

moskoweb:


templates/openclaw/index.ts
```diff
@@ -1,11 +1,15 @@
-import { Output, randomString, Services } from "~templates-utils";
+import { Output, randomPassword, Services } from "~templates-utils";
 import { Input } from "./meta";
 
 export function generate(input: Input): Output {
   const services: Services = [];
-  const gatewayToken = input.gatewayToken || randomString(64);
+  const gatewayToken = input.gatewayToken || randomPassword();
```
Ahson-Shaikh: Why changing this, I'm not sure about the length for the randomPassword but randomString with 64 bits might be even stronger. I need a strong reason for this specific change to go through.

PR #1329, opened 2026-04-10
Added template ferron

Ahson-Shaikh:
Ferron is a fast, memory-safe web server written in Rust. This template
  deploys Ferron with the official Docker image and exposes it through your
  Easypanel domain so you can start serving web traffic quickly.


Ahson-Shaikh:
Screenshot not available.

deiucanta (approved):

PR #1398, opened 2026-05-15
Updated Authorizer to v2

Ahson-Shaikh:
**index.ts**:
  - Removed the old v1 env vars (DATABASE_TYPE, DATABASE_URL, REDIS_URL, ENV)
  - Added deploy.command that invokes ./authorizer with all required v2 CLI flags
  - Connection strings (DATABASE_URL, REDIS_URL, DOMAIN) are still set as env vars so Easypanel can substitute $(PROJECT_NAME)
   and $(EASYPANEL_DOMAIN) at runtime — the shell then expands them into the flags
  - Secrets (--jwt-secret, --admin-secret, --client-id, --client-secret) are generated with randomString and embedded directly
  - Added --allowed-origins to fix the CSRF warning too

  **meta.yaml:** 1.4.4 → 2.2.1

### DIFF
--- a/templates/authorizer/index.ts
+++ b/templates/authorizer/index.ts
@@ -1,21 +1,24 @@
-import { Output, randomPassword, Services } from "~templates-utils";
+import {
+  Output,
+  randomPassword,
+  randomString,
+  Services,
+} from "~templates-utils";
 import { Input } from "./meta";
 
 export function generate(input: Input): Output {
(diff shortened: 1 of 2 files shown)


deiucanta (approved):
Reviewed locally: simulated merge with current main, inspected Authorizer diff, ran npm run build-templates.

PR #1401, opened 2026-05-18
Updated to Major Version Twenty CRM

Ahson-Shaikh:
- Updated from v1 to v2


deiucanta (approved):
Reviewed locally: simulated merge with current main, inspected Twenty CRM diff/assets, ran npm run build-templates.

PR #1443, opened 2026-06-18
Added Template Glauth

Ahson-Shaikh:
GLAuth is a lightweight, single-binary LDAP server that can be used as a
  central authentication source for self-hosted services. It supports a
  config-file backend where users and groups are defined in a TOML file, as
  well as plugin-based backends for SQLite, MySQL, and PostgreSQL. It also
  exposes a REST API for programmatic user management.


Ahson-Shaikh:
Removed the screenshot. GLAuth is a headless LDAP server daemon — it exposes no web UI at all.
  There is no browser view to capture that represents "the app as a user sees it after deploying."
  The only interfaces are the LDAP port (3893) and the REST API, neither of which produce a
  meaningful screenshot. Happy to add one if the project adds a UI in a future release.

PR #1444, opened 2026-06-18
Added Template SolidInvoice

Ahson-Shaikh:
SolidInvoice is a self-hosted invoicing and billing platform built with
  Symfony and Vue.js. Create and send professional invoices and quotes, track
  payments, manage clients and contacts, and accept online payments through
  multiple payment gateways. It ships as a single container with an embedded
  FrankenPHP server and uses MySQL for data storage.


deiucanta (approved):
Reviewed against the template checklist.

PR #1448, opened 2026-06-23
Added Template Mini QR

Ahson-Shaikh:
MiniQR is a lightweight, self-hostable QR code generator built with Vue 3
  and Vite. It lets you create fully customised QR codes — changing dot shapes,
  colours, corner styles, embedded logos, and decorative frames — and export
  them as PNG or SVG. Presets, batch export, and a clean single-page interface
  make it practical for both personal and branded use.


deiucanta (approved):

PR #1449, opened 2026-06-23
Added Template GoMFT

Ahson-Shaikh:
GoMFT is an open-source Go-based Managed File Transfer platform with a
  built-in web UI. It lets you define transfer configurations, schedule jobs
  with cron expressions, monitor transfer history, and manage users and roles —
  all from the browser. It uses rclone under the hood, giving it support for
  dozens of storage backends including S3, SFTP, Google Drive, Azure Blob, local
  paths, and more.


deiucanta (approved):

PR #1458, opened 2026-06-26
Added Template DWeebUI

Ahson-Shaikh:
DweebUI is a free, open-source web UI for managing Docker and Podman
  containers. It provides a dynamically updating dashboard showing real-time
  CPU, RAM, network, and disk metrics alongside container cards with one-click
  start, stop, pause, restart, log, and detail controls. It also manages images,
  volumes, and networks, supports multi-user access with a permission system,
  includes an app template library compatible with Portainer, and works on
  Windows, Linux, and macOS. Built with Node.js, Express, HTMX, and EJS, the
  entire stack runs in a single container connected to the Docker socket.


deiucanta (approved):

PR #1475, opened 2026-07-08
Add Manifest template

guillaumegay13:
Adds a Manifest template that deploys the public Docker image with a PostgreSQL service on Easypanel.

The template:
- creates a Manifest app service from `manifestdotbuild/manifest:6`
- creates a PostgreSQL 16 service
- wires `DATABASE_URL` through Easypanel service DNS
- generates `BETTER_AUTH_SECRET` and `MANIFEST_ENCRYPTION_KEY`
- exposes Manifest on port `2099` with Easypanel's generated HTTPS domain
- exposes the anonymous telemetry opt-out as a template option

Validation:
- `npm run build-templates` passes on this branch.
- Local smoke test completed on Easypanel v2.32.1 running in Colima, using Easypanel's Custom template `Create From Schema` flow.
- The template created both `manifest` and `manifest-db` services successfully.
- Manifest booted with PostgreSQL migrations applied and anonymous telemetry disabled.
- `/api/v1/health` returned 200 through Easypanel HTTPS routing.
- `/api/v1/setup/status` returned `needsSetup: true` before setup.
- Initial admin creation via `/api/v1/setup/admin` returned 201.
- Better Auth email sign-in returned 200 and created a secure session cookie.
- After forcing restarts of both the PostgreSQL and Manifest services, `/api/v1/health` still returned 200, setup remained `needsSetup: false`, and the session still resolved.

Manifest repo: https://github.com/mnfst/manifest
Docs: https://manifest.build


Ahson-Shaikh:
Hey, I reviewed the Manifest template PR closely. The overall structure looks mostly fine: it adds the Manifest app service, Postgres service, port `2099`, `BIND_ADDRESS=0.0.0.0`, generated auth/encryption secrets, and `BETTER_AUTH_URL`.

Before merging, can you please fix/check these points:

1. **Remove hidden/bidirectional Unicode characters**
   GitHub is showing the “hidden or bidirectional Unicode text” warning on both `templates/manifest/index.ts` and `templates/manifest/meta.yaml`. Please re-save both files as clean UTF-8 with normal LF line endings and no hidden/bidi control characters. This should be fixed before merge because it can create hard-to-debug parsing/compile issues.

2. **Use the canonical PostgreSQL URL format**
   The template currently uses:
   `postgres://postgres:<password>@...`

   Upstream Manifest docs use:
   `postgresql://user:pass@host:5432/database`

   Please switch to `postgresql://...` unless we have confirmed `postgres://` is fully accepted by Manifest/TypeORM in this image.

4. **Confirm `BETTER_AUTH_URL` matches the actual deployed URL**
   Manifest/Better Auth requires the URL to match the browser URL exactly, otherwise login can fail with “Invalid origin.” Since the template sets:
   `BETTER_AUTH_URL=https://$(PRIMARY_DOMAIN)`

   please smoke-test login/setup on the generated Easypanel domain and confirm `/setup` works without invalid-origin errors.

5. **Consider adding `SEED_DATA=false` explicitly**
   Upstream Docker Compose sets `SEED_DATA=false`. It may default to false already, but setting it explicitly keeps first installs deterministic and avoids demo/seed data surprises.

6. **Telemetry opt-out should be exposed or documented**
   Manifest self-hosted sends anonymous usage telemetry by default and supports `MANIFEST_TELEMETRY_DISABLED=1` to disable it. Since the template description emphasizes self-hosted/control of data, we should either add this env var as configurable or mention it in the instructions.

7. **Check/remove `AUTH_DB_POOL_MAX`**
   I found `DB_POOL_MAX` documented upstream, but not `AUTH_DB_POOL_MAX`. If Manifest does not actually read `AUTH_DB_POOL_MAX`, we should remove it to avoid misleading configuration.

8. **Smoke test required before merge**
   Please deploy from this template and verify:

   * app starts cleanly
   * Postgres connection/migrations succeed
   * `/api/v1/health` returns OK
   * `/setup` opens
   * first admin account creation works
   * login works on the Easypanel domain
   * restart keeps data persisted in Postgres

9. **Screenshot is Missing**

guillaumegay13:
Thanks for the detailed review. I addressed the checklist and re-tested the template.

Changes made:
- Re-saved `index.ts` and `meta.yaml`; local hidden/bidi character scan is clean.
- Switched the database URL to canonical `postgresql://...`.
- Added `SEED_DATA=false` explicitly.
- Exposed telemetry opt-out through `MANIFEST_TELEMETRY_DISABLED`.
- Kept `AUTH_DB_POOL_MAX` after confirming Manifest reads it in `packages/backend/src/auth/auth.instance.ts`.
- Added a real `screenshot1.png` captured from the Easypanel-created Manifest deployment after login.

Smoke test completed using Easypanel v2.32.1 and the documented Custom template `Create From Schema` flow:
- Template created both Manifest and PostgreSQL services successfully.
- Manifest booted cleanly and PostgreSQL migrations ran.
- `/api/v1/health` returned 200 through Easypanel HTTPS routing with `BETTER_AUTH_URL=https://$(PRIMARY_DOMAIN)`.
- `/api/v1/setup/status` reported `needsSetup: true` before setup.
- First admin creation via `/api/v1/setup/admin` returned 201.
- Better Auth email sign-in returned 200 and created a secure session cookie.
- After forcing restarts of both PostgreSQL and Manifest services, health still returned 200, setup stayed `needsSetup: false`, and the session still resolved.

Also ran `npm run build-templates` successfully on this branch before the screenshot-only update.

PR #1479, opened 2026-07-10
Added Template Poenskelisten

Ahson-Shaikh:
Pønskelisten is a self-hosted, collaborative wishlist app for creating and
  sharing gift ideas with friends and family without ruining the surprise.
  Wishes can be claimed anonymously — other members see an item is taken, but
  the list owner never finds out who claimed it. Wishlists are organized into
  groups so the same list can be shared with multiple people, and the first
  registered user automatically becomes the admin. This template runs
  Pønskelisten with its built-in SQLite database, so there is nothing extra to
  provision.


deiucanta (approved):

PR #1480, opened 2026-07-13
Added Template HabitTrove

Ahson-Shaikh:
HabitTrove is a self-hosted habit and chore tracker that pays out virtual
  coins for completed habits, which can then be redeemed against a wishlist you
  define. It supports multiple household members with their own PINs, tracks
  streaks and completion history, and stores everything in flat JSON files
  rather than a database, so there is nothing extra to provision.


deiucanta (approved):

PR #1484, opened 2026-07-14
Added Template Erugo

Ahson-Shaikh:
Erugo is a self-hosted, open source alternative to WeTransfer for sharing
  files and folders. Drag files into the browser to generate a share link with
  an optional expiry date, download limit, and password, then track who has
  downloaded what. It stores everything in a single SQLite database by default
  and generates its own encryption key and JWT secret on first boot, so there is
  no setup step beyond deploying the container.


deiucanta (approved):

PR #1485, opened 2026-07-14
Added Template Dockerizalo

Ahson-Shaikh:
Dockerizalo is a lightweight, self-hosted deployment platform for homelabbers.
  Point it at a Git repository and it clones, builds, and runs the app as a
  container on your existing Docker host, while managing secrets, volumes,
  ports, and build/runtime logs from a web dashboard. It has no built-in reverse
  proxy for deployed apps, so pair it with your own domain routing on the
  containers it creates.


deiucanta (approved):

PR #1488, opened 2026-07-20
Added Template Instatic

Ahson-Shaikh:
Instatic is a self-hosted CMS with an integrated visual canvas editor. Design
  pages, reusable components, and content models in one Bun server backed by
  SQLite, then publish plain semantic HTML and compact CSS with no builder
  runtime shipped to visitors.


deiucanta (approved):

PR #1489, opened 2026-07-20
Added Template Kitsu

Ahson-Shaikh:
Kitsu is a production tracker for animation and VFX studios, built by CGWire.
  Organize projects, shots, and assets, track task progress across your
  pipeline, and manage preview files, all from one collaborative dashboard. This
  template deploys Kitsu's real production architecture — the Zou API, its event
  stream, and the frontend behind an Nginx gateway, backed by Postgres and Redis
  — built from CGWire's own reference Dockerfile rather than the all-in-one
  image CGWire's own docs say is for testing only.


Ahson-Shaikh:
This might not be stable for production usecase of Easypanel as there is no official docker image, but building it live on Dockerfile.

deiucanta (approved):

PR #1491, opened 2026-07-21
Added Template Convoy

Ahson-Shaikh:
Convoy is a self-hosted webhooks gateway. It receives, transforms, retries,
  and delivers webhooks between your services, with a dashboard for tracking
  every event and delivery. This template runs Convoy's real production
  architecture — separate control-plane (web) and data-plane (agent) processes
  behind a Caddy gateway, backed by Postgres and Redis — matching the project's
  own reference docker-compose rather than a single container.


deiucanta (approved):
Reviewed against the README template checklist; approved for merge.

PR #1492, opened 2026-07-21
Added Template 13ft

Ahson-Shaikh:
13ft is a self-hosted paywall bypass and "read it later" style proxy. Paste an
  article URL and it fetches the page the way a search engine crawler would see
  it, falling back to Freedium or an archive.org/archive.today snapshot if the
  direct fetch is blocked. It's fully stateless with no login of its own, so it
  is protected with HTTP Basic Auth instead.


deiucanta (approved):

PR #1494, opened 2026-07-23
Added Template GhostBoard

Ahson-Shaikh:
Ghostboard is a minimal shared scratchpad — open a board URL and anyone else
  with that URL can type alongside you in real time, no account needed.
  Everything is held in memory only and clears automatically a few minutes after
  everyone disconnects, so there's nothing to back up and nothing to lose track
  of. It has no login of its own, so this template protects it with HTTP Basic
  Auth instead.


deiucanta (approved):

PR #1495, opened 2026-07-23
Added Template OTS

Ahson-Shaikh:
Ots (One-Time-Secret) lets you share a password or other sensitive text
  through a link that can only be opened once — the secret is deleted the moment
  it's viewed. It's fully open by design (anyone with the link can read it,
  that's the whole point), so it has no login of its own.


deiucanta (approved):

PR #1497, opened 2026-07-27
Fix Odoo template crashing on startup

Ahson-Shaikh:
Odoo's official Docker entrypoint refuses to boot when the Postgres connection user is literally named "postgres", as a hardcoded security guard. The template's Postgres service and app env both used that username, so every deploy crash-looped. Switched the database user to "odoo".


ellofant:
Odoo app is online and no longer keeps crashing. 

This is ready for release.

deiucanta (approved):
Reviewed against the README template checklist; approved for merge.

PR #1501, opened 2026-07-30
Added Template Cerbos

Ahson-Shaikh:
Cerbos is an open-source authorization layer that lets you define
  context-aware access control rules for your application in simple YAML
  policies, managed through your existing Git-ops workflow. It runs as a
  stateless Policy Decision Point (PDP) that your backend services call over
  HTTP or gRPC to check whether a principal can perform an action on a resource,
  supporting everything from simple RBAC to fully dynamic, attribute-based
  rules.


deiucanta (approved):

PR #1502, opened 2026-07-30
Added Template CopyParty

Ahson-Shaikh:
Copyparty turns a server into a file server with resumable uploads and
  downloads, accessible from any web browser. It supports HTTP(S), WebDAV,
  FTP(S), SFTP and TFTP in a single lightweight process, with per-folder,
  per-user permissions, thumbnails, a media player, zip/tar folder downloads,
  full-text and tag search, a markdown viewer and editor, and RSS/OPDS feeds. It
  is built for turning almost any device into a capable, self-hosted alternative
  to cloud file-sharing services.


deiucanta (approved):

PR #1524, opened 2026-08-13
Add Storyteller template

Ahson-Shaikh:
Adds Storyteller, a self-hosted platform for creating and reading ebooks with synced audiobook narration. Single container, SQLite storage on a persistent volume, no separate database service needed.

Note: Storyteller's registry only publishes per-commit image tags rather than semantic versions, so this template tracks the "latest" tag as documented by the project itself.


Ahson-Shaikh:
Closing — not ready yet, want to review the template locally first before opening a PR.

PR #1526, opened 2026-08-13
Add Riven template

Ahson-Shaikh:
Adds Riven, a media watchlist scraper and organizer for Plex/Trakt/Overseerr, backed by PostgreSQL.

Important known limitation: Riven's core streaming feature relies on an rclone-based FUSE mount to expose debrid content to a media server, which requires /dev/fuse device access and privileged mount propagation. Easypanel's app services don't expose either, so that mount will fail to start. This template is best treated as a starting point for the scraping/indexing pipeline rather than a fully working streaming setup, documented clearly in the template's instructions.


Ahson-Shaikh:
Closing — not ready yet, want to review the template locally first before opening a PR.

PR #1570, opened 2026-09-14
Bump 8mblocal to v142

Ahson-Shaikh:
Updates the 8mblocal template's default image from v137 to v142, the latest published tag.


deiucanta (approved):

PR #1573, opened 2026-09-15
Added Template Minetune

JuniorCarlini:
Minetune is a Minecraft Java server stack with its own web panel in English, Portuguese and Spanish. Pick Paper, Purpur, Fabric, NeoForge or Vanilla and the version from the official lists, manage players, game rules, plugins and mods from Modrinth and multiple worlds, and keep incremental, encrypted restic backups on local disk, Cloudflare R2 or any S3 storage. On offline-mode servers, every player creates a password for their name in a window inside the game, with no plugin or mod.

### Notes for review

- **Compose service**: the stack has five containers: the Minecraft server (itzg-based, Java 25), the web panel, a lightweight login gate that owns the game port, restic backups and a Docker socket proxy.
- **Docker socket**: the panel starts, stops and restarts the server through `wollomatic/socket-proxy`, which mounts `/var/run/docker.sock` read-only and only allows listing/inspecting containers plus start/stop/restart. Nothing can create containers.
- **Ports**: TCP 25565 (game) and UDP 19132 (optional Bedrock crossplay) are published, so Easypanel shows the expected `ports` warning. The panel is served through the Easypanel domain on `panel:8080`.
- **Data**: world, config and backups live in named volumes, so saving the Compose source or redeploying keeps the world.
- **Images**: pinned to `0.3.3` (`ghcr.io/juniorcarlini/minetune-*`), public and multi-arch (amd64 and arm64).
- **Secrets**: all four passwords are generated with `randomPassword()`.

### Tested

On an Easypanel install (Docker 29.8, arm64): deploy from the inline Compose source, domain routing to the panel, panel login with the generated password, joining the game through the gate from another machine, and saving the source plus redeploying without losing the world.


JuniorCarlini:
Thanks for the review. All points are addressed in Minetune 0.3.4, and the template now points to it:

- **Versions older than 1.21.6**: the gate no longer blocks them. Those versions have no dialog to ask for the password, so the gate just forwards the connection, the server skips the Velocity forwarding it would otherwise require, and the panel warns that there is no per-name password on that version.
- **Java 25 vs. legacy versions**: the panel now reads the JVM version inside the server container (`JAVA_VERSION`). When the image can't be swapped, as with this template, it refuses versions that need another Java and explains why, both in the version picker and when saving settings or creating a world.
- **Memory limit below the default heap**: on the first start the server reads the container memory limit and sizes the initial heap to leave about 1 GB free (for example, a `4g` limit gets a 3G heap and `2g` gets 1G), so small limits boot.
- **Wording**: the password prompt is described as offline-mode, Java-only and 1.21.6+, Bedrock players are pointed to Floodgate, and `GitHub` is capitalized.

Checked on a build of the server image: with the gate enabled, a Paper 1.21.5 server starts with Velocity forwarding disabled, and a container with a 2 GB limit starts with a 1G heap. The gate forwarding for old protocols is covered by a new test.

PR #1574, opened 2026-09-15
Update Borg UI template per maintainer feedback (image bump, links, screenshot)

Ahson-Shaikh:
Follow-up to the maintainer's review on our Easypanel docs PR (karanhudia/borg-ui#1067):

- Bump default image from `2.0.8` to `2.2.6` (current stable)
- Replace `links:` with the official GitHub repo and `https://docs.borgui.com` (dropping the stale BorgBackup/Docker Hub links)
- Swap the screenshot for a current dashboard capture from the project's own `screenshots/readme/` directory, since the old one was from the first template release


karanhudia:
Thanks for the quick turnaround, the screenshot and links look right. A few more while you are in there, all in index.ts and meta.yaml:

1. assets/logo.png is the BorgBackup project's mascot, not the Borg UI logo. Ours is https://raw.githubusercontent.com/karanhudia/borg-ui/main/frontend/public/logo-rounded.png, the same file the Unraid and Portainer templates use. It also feeds the page's social preview image, so it shows up on every shared link.
2. Add REDIS_HOST=disabled to the env. Without it the app tries a local Redis on every start, logs a failure and falls back to its in-memory cache. Same result, cleaner logs.
3. instructions is null. Worth saying there: first login is admin / admin123, change it on first use; and that archive mounting and SSHFS pull backups need a FUSE device the template does not add.
4. Optional: the / to /local bind is fixed. A hostPath input defaulting to / would let people limit what the container can see, the way our Portainer template defaults to /home.

If you would like the six feature blocks to describe what Borg UI actually adds on top of Borg (remote machines over SSH, scheduled restore checks, notification channels, Borg 2, cloud mirrors), say so and I will send copy.

Ahson-Shaikh:
Thanks for the detailed follow-up, all four addressed:

1. Swapped the logo for the correct one from `frontend/public/logo-rounded.png`.
2. Added `REDIS_HOST=disabled` to the env.
3. Filled in `instructions` with the default admin/admin123 login (change on first use) and the FUSE limitation for archive mounting/SSHFS.
4. Made the `/local` bind mount configurable via a new `localPath` field, defaulting to `/` (same pattern as your Portainer template).

On the six feature blocks — yes please, send the copy over and I'll get it updated. Happy to open a fresh PR for that with an updated screenshot once we settle the wording, per your suggestion.

deiucanta (approved):

PR #1578, opened 2026-09-16
fix: bump atomic-server image, correct GitHub link

Ahson-Shaikh:
The `atomic-server` template was pinned to `joepmeneer/atomic-server:v0.38.0`, which is stale and points at an abandoned fork (raised by the upstream maintainer on the outreach PR: https://github.com/ontola/atomic-server/pull/1422#issuecomment-5697208449).

The active repo is `ontola/atomic-server` (docs mirror: `atomicdata-dev/atomic-server`). Their CI only publishes a rolling `develop` tag to `ghcr.io/ontola/atomic-server` (no versioned tags exist on the registry), so this pins to the current `develop` digest by SHA rather than a floating tag, per this repo's pinning convention. That build is ahead of the v0.40.3 release, which shipped fixes for:
- SSRF via outbound fetches (`/bookmark`, `/import`)
- an invite-reuse race on the single-use `/setup` invite
- path traversal via `internalId` on `/download`
- stored XSS via uploaded files rendering inline

Also corrected the `links.Github` entry from `joepmeneer/atomic-server` to `ontola/atomic-server`.

Deploy-tested end to end against the new pinned digest - booted clean (HTTP 200) with the existing `atomic-storage` volume and env config untouched.


deiucanta (approved):

PR #1580, opened 2026-09-16
fix: use official gchq/cyberchef image instead of third-party mirror

Ahson-Shaikh:
Found while doing outreach for the CyberChef template: our template used the third-party `mpepping/cyberchef:v10.24.0` mirror instead of the official `ghcr.io/gchq/cyberchef` image, which is published and versioned (currently `11.4.0`, ours was 5 minor versions behind at 10.24.0).

Switched to `ghcr.io/gchq/cyberchef:11.4.0` and updated the port from 8000 to 8080 to match (per their README's docker run example). Deploy-tested - booted clean.


deiucanta (approved):

PR #1581, opened 2026-09-16
fix: set ALLOWED_ORIGINS for dataline to prevent CORS errors

Ahson-Shaikh:
Found while doing outreach for the DataLine template. Their README explicitly calls out: "The one thing you must configure when deploying DataLine to a custom domain is CORS allowed origins" - it defaults to `localhost` only.

Our template wasn't setting this, so the deployed UI would hit CORS errors talking to its own backend on any real domain. Added `ALLOWED_ORIGINS=https://$(PRIMARY_DOMAIN)`.

Deploy-tested - confirmed the app loads and successfully calls its own API (reaches the OpenAI key config screen) with the fix in place.


deiucanta (approved):

PR #1582, opened 2026-09-16
fix: use deployed domain for docmost APP_URL instead of hardcoded localhost

Ahson-Shaikh:
Found while doing outreach for the Docmost template. Their docs explicitly warn `APP_URL` must be the real domain ("should be the domain you can access docmost from"), and that real-time collaboration relies on WebSockets working correctly off this value.

Our template hardcoded `APP_URL=http://localhost:3000` regardless of the deployed domain. Fixed to use `https://$(PRIMARY_DOMAIN)`.

Deploy-tested - confirmed the app boots and the setup/register page loads correctly with the fix.


deiucanta (approved):

PR #1587, opened 2026-09-17
Fix DVinyl template: MongoDB auth failure

Ahson-Shaikh:
## Summary
- Fixes `MONGODB_URL` in `templates/dvinyl/index.ts` to append `?authSource=admin`

## Why
Found while testing the template for an outreach PR. Deploying the template as-is fails to boot: the app can't connect to MongoDB because the connection string points at `/dvinyl` as both the target database and (implicitly) the auth database, but the `mongo` user only exists in the `admin` database. MongoDB logs show:
```
UserNotFound: Could not find user "mongo" for db "dvinyl"
```

## Test plan
- [x] Deployed the unfixed template via Easypanel demo project — confirmed the auth failure in mongo logs
- [x] Applied the fix, redeployed fresh, confirmed the app boots and responds (302 redirect to login)


Ahson-Shaikh:
Closing this — I made an error. This assumed the DVinyl template already existed on easypanel-io/templates, but it turns out DVinyl (along with a couple other templates I was testing) only exists in a local, unpushed working copy and was never actually added to this repo. This PR would have only added templates/dvinyl/index.ts on its own, without meta.yaml or assets, which isn't a usable template. Will follow up separately once the template is properly added.

